PR #19, opened 2022-11-11
fix dataset.stat() bug for empty list

zzachw:
added some validation assertions

### DIFF
--- a/pyhealth/datasets/base_dataset.py
+++ b/pyhealth/datasets/base_dataset.py
@@ -10,7 +10,8 @@ from tqdm import tqdm
 
 from pyhealth.data import Patient, Event
 from pyhealth.datasets.utils import MODULE_CACHE_PATH
-from pyhealth.datasets.utils import hash_str, list_nested_level, is_homo_list
+from pyhealth.datasets.utils import hash_str
+from pyhealth.datasets.utils import list_nested_level, is_homo_list, flatten_list
 from pyhealth.medcode import CrossMap
 from pyhealth.utils import load_pickle, save_pickle
 
@@ -356,9 +357,7 @@ class BaseDataset(ABC, Dataset):
                 # 1 level nested list
                 if level == 1:
                     # a list of values of the same type
-                    # sum() flattens the nested list
-                    # e.g, [[1, 2], [3, 4]] -> [1, 2, 3, 4]
-                    check = is_homo_list(sum([s[key] for s in samples], []))
+                    check = is_homo_list(flatten_list([s[key] for s in samples]))
                     assert check, \
                         f"Key {key} has mixed types in the nested list within samples"
                 # 2 level nested list
@@ -370,8 +369,9 @@ class BaseDataset(ABC, Dataset):
                     assert all(check), \
                         f"Key {key} has mixed nested list levels within samples"
                     # a list of list of values of the same type
-                    # sum() flattens the nested list
-                    check = is_homo_list(sum([l for s in samples for l in s[key]], []))
+                    check = is_homo_list(
+                        flatten_list([l for s in samples for l in s[key]])
+                    )
                     assert check, \
                         f"Key {key} has mixed types in the nested list within samples"
 
@@ -459,8 +459,7 @@ class BaseDataset(ABC, Dataset):
                     continue
                 # a list of lists of values
                 elif type(sample[key][0]) == list:
-                    # sum() flattens the nested list
-                    tokens.extend(sum(sample[key], []))
+                    tokens.extend(flatten_list(sample[key]))
                 # a list of values
                 else:
                     tokens.extend(sample[key])
@@ -568,8 +567,9 @@ class BaseDataset(ABC, Dataset):
                 nested = [isinstance(e, list) for s in self.samples for e in s[key]]
                 # key's feature is a list of lists
                 if any(nested):
-                    # sum() flattens the nested list
-                    num_events = [len(sum(sample[key], [])) for sample in self.samples]
+                    num_events = [
+                        len(flatten_list(sample[key])) for sample in self.samples
+                    ]
                 # key's feature is a list of values
                 else:
                     num_events = [len(sample[key]) for sample in self.samples]
--- a/pyhealth/datasets/base_dataset.py
+++ b/pyhealth/datasets/base_dataset.py
@@ -1,15 +1,14 @@
 import logging
-import os
 from abc import ABC, abstractmethod
 from collections import Counter
 from copy import deepcopy
-from typing import Optional, List, Dict, Callable, Tuple, Union
+from typing import Dict, Callable, Tuple, Union
 
 from torch.utils.data import Dataset
 from tqdm import tqdm
 
 from pyhealth.data import Patient, Event
-from pyhealth.datasets.utils import MODULE_CACHE_PATH, hash_str
+from pyhealth.datasets.utils import *
 from pyhealth.medcode import CrossMap
 from pyhealth.utils import load_pickle, save_pickle
 
@@ -309,6 +308,42 @@ class BaseDataset(ABC, Dataset):
                 self.patients.items(), desc=f"Generating samples for {self.task}"
         ):
             samples.extend(self.task_fn(patient))
+
+        # validate the samples
+        assert all(isinstance(s, dict) for s in samples), "Each sample should be a dict"
+        keys = samples[0].keys()
+        assert all(set(s.keys()) == set(keys) for s in samples), \
+            "All samples should have the same keys"
+        assert "patient_id" in keys, "patient_id should be in the keys"
+        assert "visit_id" in keys, "visit_id should be in the keys"
+        for key in keys:
+            # either a single value, a list of values, or a list of list of values
+            check = is_homo_list([s[key] for s in samples])
+            assert check, f"Key {key} has mixed types across samples"
+            type_ = type(samples[0][key])
+            # 1 or 2 level nested list
+            if type_ == list:
+                levels = set([list_nested_level(s[key]) for s in samples])
+                assert len(levels) == 1, \
+                    f"Key {key} has mixed nested list levels across samples"
+                level = levels.pop()
+                assert level in [1, 2], \
+                    f"Key {key} has unsupported nested list level across samples"
+                # 1 level nested list
+                if level == 1:
+                    check = [is_homo_list(s[key]) for s in samples]
+                    assert all(check), \
+                        f"Key {key} has mixed types in the nested list within samples"
+                # 2 level nested list
+                else:
+                    check = [is_homo_list(s[key]) for s in samples]
+                    assert all(check), \
+                        f"Key {key} has mixed nested list levels within samples"
+                    check = [is_homo_list(l) for s in samples for l in s[key]]
+                    assert all(check), \
+                        f"Key {key} has mixed types in the nested list within samples"
+
+        # set the samples
         self.samples = samples
         self.patient_to_index = self._index_patient()
         self.visit_to_index = self._index_visit()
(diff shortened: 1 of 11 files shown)


pyhealth/datasets/base_dataset.py
```diff
@@ -1,15 +1,14 @@
...
 from collections import Counter
 from copy import deepcopy
-from typing import Optional, List, Dict, Callable, Tuple, Union
+from typing import Dict, Callable, Tuple, Union
 
 from torch.utils.data import Dataset
 from tqdm import tqdm
 
 from pyhealth.data import Patient, Event
-from pyhealth.datasets.utils import MODULE_CACHE_PATH, hash_str
```
ycq091044, line 12: It is still good to explicitly write down what we are importing, otherwise, we may have to look into `pyhealth.datasets.utils` and check what tools are available.

pyhealth/datasets/base_dataset.py
```diff
@@ -309,6 +308,42 @@ def set_task(
...
+        # validate the samples
+        assert all(isinstance(s, dict) for s in samples), "Each sample should be a dict"
+        keys = samples[0].keys()
+        assert all(set(s.keys()) == set(keys) for s in samples), \
+            "All samples should have the same keys"
+        assert "patient_id" in keys, "patient_id should be in the keys"
+        assert "visit_id" in keys, "visit_id should be in the keys"
+        for key in keys:
+            # either a single value, a list of values, or a list of list of values
+            check = is_homo_list([s[key] for s in samples])
```
ycq091044, line 342: I just jumped to `is_homo_list`. It would be good to have "Example" for this function.

pyhealth/datasets/base_dataset.py
```diff
@@ -309,6 +308,42 @@ def set_task(
...
+            assert check, f"Key {key} has mixed types across samples"
+            type_ = type(samples[0][key])
+            # 1 or 2 level nested list
+            if type_ == list:
+                levels = set([list_nested_level(s[key]) for s in samples])
```
ycq091044, line 351: I just jumped to `list_nested_level`. It would be good to have "Example" for this function.

pyhealth/datasets/base_dataset.py
```diff
@@ -386,17 +422,22 @@ def get_all_tokens(
             raise ValueError("Please set task first.")
         tokens = []
         for sample in self.samples:
+            if type(sample[key]) == list:
+                if len(sample[key]) == 0:
+                    continue
+                # a list of lists of values
+                elif type(sample[key][0]) == list:
+                    tokens.extend(sum(sample[key], []))
```
ycq091044: I am not sure whether this `sum` operation works for concatenating two lists. I have tried an example, which gives me an error.
```
>>> sum([1,2,3], [])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate list (not "int") to list
```

pyhealth/datasets/base_dataset.py
```diff
@@ -386,17 +422,22 @@ def get_all_tokens(
...
             # single value
-            if type(sample[key]) in [bool, int, str]:
-                tokens.append(sample[key])
-            # a list of values
-            elif type(sample[key][0]) in [bool, int, str]:
-                tokens.extend(sample[key])
-            # a list of lists of values
-            elif type(sample[key][0]) == list:
-                tokens.extend(sum(sample[key], []))
             else:
```
ycq091044, line 467: Here, we only consider the list and list of list cases. However, maybe the `is_homo_list` function does not need that recursion.

pyhealth/datasets/base_dataset.py
```diff
@@ -386,17 +422,22 @@ def get_all_tokens(
...
-                raise ValueError(f"Unknown type of {key}: {type(sample[key])}")
+                tokens.append(sample[key])
+        types = set([type(t) for t in tokens])
+        assert len(types) == 1, f"{key} tokens have mixed types"
+        assert types.pop() in [int, float, str, bool], \
```
ycq091044, line 471: right now, I think we only use the `str` value?

pyhealth/datasets/base_dataset.py
```diff
@@ -482,29 +524,32 @@ def task_stat(self) -> str:
...
-                num_events = [1 for sample in self.samples]
-            # a list of values
-            elif type(self.samples[0][key][0]) in [bool, int, str]:
-                num_events = [len(sample[key]) for sample in self.samples]
-            # a list of lists of values
-            elif type(self.samples[0][key][0]) == list:
-                num_events = [len(sum(sample[key], [])) for sample in self.samples]
+            # list
+            if type(self.samples[0][key]) == list:
+                nested = [isinstance(e, list) for s in self.samples for e in s[key]]
```
ycq091044, line 567: Should we improve this part?

pyhealth/datasets/base_dataset.py
```diff
@@ -482,29 +524,32 @@ def task_stat(self) -> str:
...
+                if any(nested):
+                    num_events = [len(sum(sample[key], [])) for sample in self.samples]
+                else:
```
ycq091044, line 574: `else` may mean that samples have different nested levels.

pyhealth/datasets/base_dataset.py
```diff
@@ -482,29 +524,32 @@ def task_stat(self) -> str:
...
-            elif type(self.samples[0][key][0]) in [bool, int, str]:
-                num_events = [len(sample[key]) for sample in self.samples]
-            # a list of lists of values
-            elif type(self.samples[0][key][0]) == list:
-                num_events = [len(sum(sample[key], [])) for sample in self.samples]
+            # list
+            if type(self.samples[0][key]) == list:
+                nested = [isinstance(e, list) for s in self.samples for e in s[key]]
+                if any(nested):
+                    num_events = [len(sum(sample[key], [])) for sample in self.samples]
```
ycq091044: again, the `sum(list, list)` may cause error.

pyhealth/datasets/utils.py
```diff
@@ -32,6 +33,36 @@ def strptime(s: str) -> Optional[datetime]:
...
+def list_nested_level(l: List) -> int:
+    """Gets the nested level of a list.
+
+    Args:
+        l: List, the list to be checked.
+
+    Returns:
+        int, the nested level of the list.
+    """
+    if not isinstance(l, list):
```
ycq091044, line 72: add Example

pyhealth/datasets/utils.py
```diff
@@ -32,6 +33,36 @@ def strptime(s: str) -> Optional[datetime]:
...
+
+def is_homo_list(l: List) -> bool:
+    """Checks if a list is homogeneous.
+
+    Args:
+        l: the list to be checked.
+
+    Returns:
+        bool, True if the list is homogeneous, False otherwise.
+    """
```
ycq091044, line 97: Add Example

ycq091044 (changes requested):
some issues and suggestions, see comments.

ycq091044 (approved):
LGTM

ycq091044 (changes requested):
I just gave separate reviews for two commits. Let me use a final one to summarize: needs changes on the second commits.

ycq091044:
Right now, you separately check "list" or "list of list" structures in different functions. Should we combine the check using a common function?

pyhealth/datasets/base_dataset.py
zzachw: resolved in the latest commit

pyhealth/datasets/base_dataset.py
zzachw: resolved in the latest commit

pyhealth/datasets/base_dataset.py
zzachw: resolved in the latest commit

pyhealth/datasets/base_dataset.py
zzachw: sum() flattens the nested list, e.g. [[1, 2], [3, 4]] -> [1, 2, 3, 4]

pyhealth/datasets/base_dataset.py
zzachw: Do you mean `list_nested_level()`? I try to make `list_nested_level()` more general instead of only applying to two level nested list.

pyhealth/datasets/base_dataset.py
zzachw: Yes. I try to leave some flexibility for the future (e.g., regression task)

pyhealth/datasets/base_dataset.py
zzachw: more comments are added to explain the rationale

pyhealth/datasets/base_dataset.py
zzachw: It can either be 1 or 2 level list as we check this in the end to set_task.

pyhealth/datasets/base_dataset.py
zzachw: sum() flattens the nested list, e.g. [[1, 2], [3, 4]] -> [1, 2, 3, 4]

pyhealth/datasets/utils.py
zzachw: done

pyhealth/datasets/utils.py
zzachw: done

zzachw:
All comments should be resolved in the latest commit

zzachw:
> Right now, you separately check "list" or "list of list" structures in different functions. Should we combine the check using a common function?

I didn't figure out a way to perform the check in the end of `self.set_task()` with one common function. I have listed the things I checked in the code. Let me know if you have any ideas.

pyhealth/datasets/base_dataset.py
ycq091044: **[New Comments]** Then, should we add one more bracket and use `sum([sample[key], []])`.

However, I also tried this
```
>>> sum([[1,2,3], []])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'int' and 'list'
```

pyhealth/datasets/base_dataset.py
ycq091044: ok

pyhealth/datasets/base_dataset.py
ycq091044: **[New Comment]** I mean what if `nested=[True, False, True, False]` ? Then, in this else-loop, some are 1-level, some are 2-level?

pyhealth/datasets/base_dataset.py
zzachw: The correct usage is `sum([[1, 2], [3, 4]], start=[])`.

In the latest commit, [27ddaa7](https://github.com/sunlabuiuc/PyHealth/commit/27ddaa7254ba94fea47b9a27d0cfcefdbdc21be9), I wrapped this into a function `flatten_list()` for easier understanding.

pyhealth/datasets/base_dataset.py
zzachw: https://github.com/sunlabuiuc/PyHealth/blob/27ddaa7254ba94fea47b9a27d0cfcefdbdc21be9/pyhealth/datasets/base_dataset.py#L363-L370

This case is eliminated at the end of `set_task()`.

zzachw:
Wrapped `sum()` into a function flatten_list() for easier understanding.

ycq091044 (approved):
Finally, LGTM. Thanks zhenbang for your hard work!

PR #23, opened 2022-11-16
Fix leaderboard and add advanced tutorial 4

ycq091044:


pyhealth/datasets/eicu.py
```diff
@@ -73,27 +73,6 @@ class eICUDataset(BaseDataset):
...
-            self,
-            root: str,
-            tables: List[str],
-            code_mapping: Optional[Dict[str, Union[str, Tuple[str, Dict]]]] = None,
-            dev=False,
-            refresh_cache=False,
-    ):
-        # store a mapping from visit_id to patient_id
-        # will be used to parse clinical tables as they only contain visit_id
-        self.visit_id_to_patient_id: Dict[str, str] = {}
```
zzachw, line 86: We have to keep the init for eicu, as it has a special parameter self.visit_id_to_patient_id

pyhealth/datasets/base_dataset.py
```diff
@@ -76,7 +76,6 @@ class BaseDataset(ABC, Dataset):
 
     def __init__(
             self,
-            dataset_name: str,
```
zzachw, line 79: I think we should still keep this arg for flexibility. It this arg is not provided, we will use self.__class__.__name__

zzachw (changes requested):

pyhealth/datasets/base_dataset.py
zzachw: E.g., `dataset_name: Optional[str] = None`

pyhealth/datasets/eicu.py
zzachw: E.g.
```
def __init__(self, **kwargs):
  # store a mapping from visit_id to patient_id
  # will be used to parse clinical tables as they only contain visit_id
  self.visit_id_to_patient_id: Dict[str, str] = {}
  super(eICUDataset, self).__init__(**kwargs)
```

zzachw (approved):
LGTM

PR #27, opened 2022-11-17
support float or input sequences input

ycq091044:


zzachw (changes requested):
I think you should also update transformer.event_level_forward() to support vector input

ycq091044:
> I think you should also update transformer.event_level_forward() to support vector input

We can remove the event function. Or build a transformer model and let the user choose to run on visit or on event or on visit and event.

zzachw (approved):
LGTM

PR #28, opened 2022-11-20
fix train printing format

ycq091044:


zzachw (approved):
LGTM. I will fix the logging issue on Colab in a separate PR later

PR #39, opened 2022-12-07
Fix logging problem

zzachw:


pyhealth/__init__.py
```diff
@@ -10,7 +10,10 @@
     os.makedirs(BASE_CACHE_PATH)
 
 # logging
-logger = logging.getLogger()
+logger = logging.getLogger(__name__)
```
ycq091044, line 13: Could I learn what does this change mean?

ycq091044 (approved):
LGTM. Will this replacement from "logging" to "logger" solve the printing problem? I would be interested to know why.

pyhealth/__init__.py
zzachw: `logging.getLogger()` gives you the root logger (shared by all applications using the logging module) while `logging.getLogger(__name__)` gives you a child logger (whose name is "pyhealth").

I changed this because we don't want to modify the behavior of the root logger which may be used outside of our package.

FYI: a good reference https://colab.research.google.com/github/aviadr1/learn-advanced-python/blob/master/content/15_logging/logging_in_python.ipynb

PR #60, opened 2023-01-05
allow special tokens to be passed through add_feature_transform_layer

zlin7:
I noticed that some models use get_feature_tokenizers and some switched to add_feature_transform_layer, but the latter IMO should take special_tokens as an argument as well.


ycq091044:
discussed offline. Please add the changes to `doc/logs.rst`

PR #61, opened 2023-01-09
Add model deepr

zlin7:
https://arxiv.org/abs/1607.07519

### DIFF
--- a/docs/api/models.rst
+++ b/docs/api/models.rst
@@ -15,4 +15,5 @@ We implement the following models for supporting multiple healthcare predictive 
     models/pyhealth.models.GAMENet
     models/pyhealth.models.MICRON
     models/pyhealth.models.SafeDrug
+    models/pyhealth.models.Deepr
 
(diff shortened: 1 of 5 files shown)


ycq091044:
Nice job @zlin7 . Please register the deeper model into `docs/api/model/` and `docs/api/models.rst` following the example. Also, could you please summarize the newest changes (I believe you participate in all commits) and add them into `docs/log.rst`.

ycq091044:
Nit: below the argument description, it would be good to add examples for the layer (better to add in the model as well), e.g.,

```bash
Examples:
        >>> from pyhealth.models import CNNLayer
        >>> input = torch.randn(3, 128, 5)  # [batch size, sequence len, input_size]
        >>> layer = CNNLayer(5, 64)
        >>> outputs, last_outputs = layer(input)
        >>> outputs.shape
        torch.Size([3, 128, 64])
        >>> last_outputs.shape
        torch.Size([3, 64])
```

Active your feature branch in readthedocs website. Once the changes show up, you can merge into the develop.

PR #78, opened 2023-02-01
Add unittest of medcode and tokenizer

pat-jj:
Added unittest of medcode and tokenizer, please check the code.


ycq091044:
LGTM. Some comments:
1. Could you add your updates to doc/logs.rst?
2. Could you add the examples you use (in unit test tokenizer) for this page https://pyhealth.readthedocs.io/en/latest/api/tokenizer.html. I think that since unittest file will contain examples anyway, let us also update the examples to the webpage (if there is no example before).

pat-jj:
> 

Sure, will do soon.

PR #108, opened 2023-02-20
support 6 new EHR models from covid benchmark

ycq091044:
- redirected from junyi's PR on #106 
- new website: https://pyhealth.readthedocs.io/en/covid-benchmark/


ycq091044:
@v1xerunt For agent.py, I get the following error if I do not use the "static_key".

image

Ok, I find this also does not run through

image

also this:

image

ycq091044:
@v1xerunt 

1. Please start from this branch and modify the code.
2. I have check the adacare, agent, and concare models, but not the other three. Right now, the adacare model works for me. But the agent and concare models do not work. As I have shown above, the AgentLayer and ConcareLayer examples you provided do not work.
3. Minor: you may use PEP8 code formatted before sending the PR

PR #111, opened 2023-02-22
Uncertainty quantification modules

zlin7:


zlin7:
Before making this draft ready"
1. Remove dependency on: `ipdb`, `persist_to_disk`
2. Reorganize the metrics

zlin7:
Added methods; KCal, Temperature Scaling, Histogram Binning, SCRIB, LABEL (see `pyhealth/uq/README.md`)

Added evaluation metrics for calibration and prediction set evaluation. 

TODO: Remove the test_class.py file once this is done.

pyhealth/calib/utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+class LogLoss(torch.nn.Module):
+    """Cross entropy, but takes in the probability instead of the logits"""
+    reduction: str
+    def __init__(self, weight: Optional[Tensor] = None,  ignore_index: int = -100, reduction: str = 'mean', clip=1e-10) -> None:
+        super(LogLoss, self).__init__()
+        self.register_buffer('weight', weight)
+        self.ignore_index = ignore_index
+        self.reduction = reduction
+        self.clip = clip
```
ycq091044, line 33: what is Line27 used for?

pyhealth/trainer.py
```diff
@@ -271,9 +272,16 @@ def inference(self, dataloader) -> Dict[str, float]:
                 loss_all.append(loss.item())
                 y_true_all.append(y_true)
                 y_prob_all.append(y_prob)
+                if additional_outputs is not None:
+                    for key in additional_outputs.keys():
+                        additional_outputs[key].append(output[key].cpu().numpy())
```
ycq091044, line 277: Should we add `if key in output` in the middle?

pyhealth/metrics/calibration.py
```diff
@@ -0,0 +1,118 @@
...
+    return _ECE_confidence(df, bins, adaptive)[1]
+
+def ECE_confidence_binary(prob:np.ndarray, label:np.ndarray, bins=20, adaptive=False):
+    df = pd.DataFrame({'acc': label[:,0], 'conf': prob[:,0]})
+    return _ECE_confidence(df, bins, adaptive)[1]
+
+def brier_top1(prob:np.ndarray, label:np.ndarray):
+    conf = prob.max(1)
+    acc = (label == np.argmax(prob, 1)).astype(int)
+    return np.mean(np.square(conf - acc))
```
ycq091044: We should present some examples to show on readthedocs somehow.

pyhealth/metrics/prediction_set.py
```diff
@@ -0,0 +1,64 @@
...
+    truth_pred = y_pred[np.arange(len(y_true)), y_true]
+
+    return 1 - np.mean(truth_pred)
+
+def missrate_certain_overall(y_pred:np.ndarray, y_true:np.ndarray):
+    """Miscoverage rate for the true label for unrejected samples. Only for multiclass"""
+    assert len(y_true.shape) == 1
+    truth_pred = y_pred[np.arange(len(y_true)), y_true]
+    truth_pred = truth_pred[y_pred.sum(1) == 1]
+    return 1 - np.mean(truth_pred)
```
ycq091044, line 116: The same, we should provide some easy-to-understand examples somehow.

pyhealth/calib/utils.py
zlin7: reduction is just specifying whether we should reduce the loss by mean/sum etc.

I was trying to follow https://pytorch.org/docs/1.10/_modules/torch/nn/modules/loss.html#CrossEntropyLoss in how they write loss

pyhealth/trainer.py
zlin7: I'm leaning to not but am OK with either. I think `additional_outputs` is used only for users/developers who know what they are doing, and if we don't trigger errors it could be hard to debug.

pyhealth/metrics/prediction_set.py
zlin7: @ycq091044 Just added more examples to both.

pyhealth/calib/utils.py
ycq091044: ok, I thought you already define "reduction: str" in the function argument, and it would be unnecessary. But never mind.

pyhealth/trainer.py
ycq091044: That makes sense. But this option is exposed to the users, they might want to add something. I saw you write "(only if requested)". Ok, let us leave it here.

ycq091044:
One last question: In metrics, what is the relationship between the prediction_set, calibration and the previous binary, multilabel, multiclass? 

I find that the multiclass and binary can also compute some calibration or prediction_set metrics. I feel like calibration and prediction_set are not at the same level as binary/multilabel/multiclass. Could we merge them into the previous three metrics category and not expose the calibration.py and prediction_set.py to users>

ycq091044:
Thanks @zlin7 . I have some tiny nits to ask/suggest:
- Are we going to name the module "Calibration and Uncertainty Quantification"? I thought we agree on "calibration". I am fine with both, but just want to make sure you create a proper name.
- Would you mind adding some outputs in the web examples. For instance, you have the following for `pyhealth.calib.calibration.KCal`. It might be good to let people know what is the output of last line. User might be curious what is the style? what is the range? is it a scalar or a list of something?
```python
from pyhealth.models import SparcNet
from pyhealth.tasks import sleep_staging_isruc_fn
sleep_ds = ISRUCDataset("/srv/scratch1/data/ISRUC-I").set_task(sleep_staging_isruc_fn)
train_data, val_data, test_data = split_by_patient(sleep_ds, [0.6, 0.2, 0.2])
model = SparcNet(dataset=sleep_staging_ds, feature_keys=["signal"],
    label_key="label", mode="multiclass")
# ... Train the model here ...
# Calibrate
cal_model = uq.KCal(model)
cal_model.calibrate(cal_dataset=val_data)
# Alternatively, you could re-fit the reprojection:
# cal_model.calibrate(cal_dataset=val_data, train_dataset=train_data)
# Evaluate
from pyhealth.trainer import Trainer
test_dl = get_dataloader(test_data, batch_size=32, shuffle=False)
print(Trainer(model=cal_model).evaluate(test_dl))

```

Other parts look great to me!

PR #121, opened 2023-03-03
Add new model MoleRec

zengkaipeng:
Add the official implementation of Nianzu Yang et al. MoleRec: Combinatorial Drug Recommendation with Substructure-Aware Molecular Representation Learning. WWW 2023.


ycq091044:
Moved to a new base, so I can test locally.

PR #125, opened 2023-03-18
Ambiguous datetime usage on dateutils.parser.parse causes incorrect timestamps in eICU

BPDanek:
Since in eICU the admit time is only provided a year, the dateutil_parse function was receiving a parameter like "2014", while its output was of shape "YYYY-MM-DD". The "MM-DD" were not provided, and inferred to be _the current day_. Downstream timestamps are computed based on the offset of the admit time, so all downstream timestamps are incorrect as a result.

The way unit tests were written, when writing the test, we would fill in hardcoded values on all the expected fields of the dataset, so we would hardcode the current day into the test. This would cause the test to pass one day, and then fail the next. 

Notice how in the test fields, the dates on all the attributes are in late February. This is not a coincidence, that is just when we were writing the tests. 

Screenshot 2023-03-17 at 1 18 43 PM

### DIFF
--- a/pyhealth/datasets/eicu.py
+++ b/pyhealth/datasets/eicu.py
@@ -7,7 +7,7 @@ from datetime import datetime
 
 from pyhealth.data import Event, Visit, Patient
 from pyhealth.datasets import BaseEHRDataset
-from pyhealth.datasets.utils import strptime
+from pyhealth.datasets.utils import strptime, padyear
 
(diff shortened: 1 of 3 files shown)


BPDanek:
test outcome: https://github.com/sunlabuiuc/PyHealth/actions/runs/4454778248

ycq091044 (approved):
LGTM.

PR #136, opened 2023-04-23
WIP - MIMIC-Extract dataset

SphtKr:
Work In Progress - Dataset reader for [MIMIC-Extract](https://github.com/MLforHealth/MIMIC_Extract) output support. This allows the HDF5 output files from their preprocessing pipeline to be read in PyHealth. An example of the dataset is [available on GCP](https://console.cloud.google.com/storage/browser/mimic_extract).

Needs a little docs work and probably a unit test (though I'd probably have to hand-roll a small file set for this), but looking for feedback before I do that: does this need any particular rework?

One area I'm not sure how to handle  is that this requires pytables to read the HDF5. Pandas already has a "soft" optional dependency on HDF5, and if you try to use this dataset without pytables installed you will get their exception saying it needs to be installed. I opted to make this formal by adding an extras definition to setup.py/setup.cfg--I mirrored what pandas is doing in v2.0 (notably not in 1.5 however), but this may not be the core developers' preference. Thoughts?

### DIFF
--- a/pyhealth/datasets/mimicextract.py
+++ b/pyhealth/datasets/mimicextract.py
@@ -140,6 +140,10 @@ class MIMICExtractDataset(BaseEHRDataset):
         # sort by admission and discharge time
         df = patients_df.reset_index().sort_values(["subject_id", "admittime", "dischtime"], ascending=True)
         # group by patient
+        #TODO: This can probably be simplified--MIMIC-Extract includes only the first ICU
+        # visit for each patient (see paper)... it is unclear whether it might be easily
+        # modified to include multiple visits however, so this may have value for customised
+        # versions of the pipeline.
         df_group = df.groupby("subject_id")
 
         # parallel unit of basic information (per patient)
--- a/pyhealth/datasets/mimicextract.py
+++ b/pyhealth/datasets/mimicextract.py
@@ -96,6 +96,24 @@ class MIMICExtractDataset(BaseEHRDataset):
                 #self._vocab_map[linksto] = df[["ITEMID"]].to_dict(orient="index")
                 for r in df.iterrows():
                     self._vocab_map[linksto][r[0].lower()] = r[1]["ITEMID"]
+
+        # reverse engineered from mimic-code concepts SQL and MIMIC-Extractt SQL...
+        self._vocab_map['interventions'] = { 
+            'vent': 467,
+            'adenosine': 4649,
+            'dobutamine': 30042,
+            'dopamine': 30043,
+            'epinephrine': 30044,
+            'isuprel': 30046,
+            'milrinone': 30125,
+            'norepinephrine': 30047,
+            'phenylephrine': 30127,
+            'vasopressin': 30051,
+            'colloid_bolus': 46729, # "Dextran" Arbitrary! No general itemid!
+            'crystalloid_bolus': 41491, # "fluid bolus"
+            'nivdurations': 468
+        }
+
                 
         super().__init__(root=root, tables=tables,
             dataset_name=dataset_name, code_mapping=code_mapping,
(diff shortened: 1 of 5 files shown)


SphtKr:
I've recently gotten this pipeline to process the MIMIC-III demo dataset. I'm not 100% sure, but I think this can be redistrbuted. The derived dataset is about 46MB, and could be used for unit tests, but I don't know if you'd want this in the PyHealth repo... it's an option though, also open to other suggestions.

ycq091044:
Thanks @SphtKr , will take a look over the weekends.

setup.cfg
```diff
@@ -1,2 +1,6 @@
 [metadata]
-description-file = README.rst
\ No newline at end of file
+description-file = README.rst
+
+[options.extras_require]
+hdf5 =
+    tables>=3.7.0 # Pandas 2.0 has an hdf5 variant... use that later?
```
ycq091044: do not worry, we restrict the pandas pacakge requirements to <2. We can ignore this part.

setup.py
```diff
@@ -47,6 +47,9 @@ def readme():
     include_package_data=True,
     install_requires=requirements,
     setup_requires=["setuptools>=38.6.0"],
+    extras_require={
+        "hdf5": ["tables>=3.7.0"], # Pandas 2.0 has an hdf5 variant... use that later?
+    },
```
ycq091044: follow the above, also ignore this part.

pyhealth/datasets/mimicextract.py
```diff
@@ -0,0 +1,475 @@
...
+        root="../data/baseline5000/grouping",
+        tables=[
+            "C",
+            "vitals_labs",
+            "interventions"
+        ],
+        #code_mapping={"NDC": "ATC"},
+        dev=True,
+        refresh_cache=True,
+    )
```
ycq091044, line 544: Could you give me a link to the MIMIC-Extract dataset? Also, could you send a pr (not draft), so that I can merge your code to pyhealth/mimic-extract for local testing. If everything is fixed, I will merge into master.

setup.cfg
ycq091044: @SphtKr

pyhealth/datasets/mimicextract.py
SphtKr: I've uploaded two datasets (different grouping options) here:
https://drive.google.com/drive/folders/1RaioXfcHYKUwyL__LdUcbVnmLXrgDUmp?usp=sharing

I've also included a copy of the `itemid_to_variable_map.csv` file from MIMIC-Extract in there for convenience.

ycq091044:
Thanks for your contribution! @SphtKr , I downloaded your data in google drive (I assume they are extracted from the mimic-iii demo dataset) and replicate your work on my local environment. The code runs smoothly while I need to use "conda install pytables" at the beginning. 

I am trying the understand how users will use this MIMICExtractDataset class (correct me if I am wrong).
1. they will first have their MIMIC-III or MIMIC-IV dataset ready.
2. they go to https://github.com/MLforHealth/MIMIC_Extract and use the tool there to extract the new tables.
> I did not try the tool but **I am curious whether the extracted tables alone are necessary to run the MIMICExtractDataset class.** Or we should put the new tables with the original files (DIAGNOSES.csv, PROCEDURES.csv, etc). After looking at the folders (grouping or nogrouping) you gave me, I assume the former. Right?

3. they go and run the MIMICExtractDataset class on the new tables. **One question here**: I did not see the patients.csv or admission.csv in your grouping or nogrouping folder. However, the codes do not give error. What is the magic?

Thanks for your continueous efforts on this PR. I would like to get some inputs on how this dataset is different from the original MIMIC-III? What is the benefit here? Are the above procedures the right way to use the MIMICExtractDataset class? It would be good we can schedule a quick zoom call to sync.

SphtKr:
@ycq091044 You've about got it. The procedures are correct, though what happens is a little involved, including setting up a PostgreSQL version of MIMIC-III as an intermediate step (I am also about to make that easier).

The [paper on MIMIC-Extract](https://arxiv.org/abs/1907.08322) gives all the detail on what MIMIC-Extract provides beyond the raw MIMIC-III data, but in a nutshell it is a cleaning/preprocessing/cohort selection tool. Besides some basic cohort selection filters (e.g. min-age), it cleans the data by unifying/converting units, extracting some data from notes, unifying ITEMIDs for the same or similar items (based on medical ontology knowledge and at one of two levels of grouping--the `grouping` directory in the Google Drive is "LEVEL2" grouping while the `nogrouping` directory is processed as "LEVEL1"), and eliminating or clamping outliers in the data. As a particular enhancement, the `interventions` table includes features representing several classes of ICU interventions, which the authors used to train models for onset/wean prediction tasks.

The MIMIC-Extract HDF5 files contain a subset of the full MIMIC-III data, including a table (HDF5 "dataset") inside `all_hourly_data.h5` with the selected cohort's patient/admissions data—this is what is processed in `parse_basic_info`—so this is how the original MIMIC-III .csv files are not required. The `C.h5` file also contains the ICD9 diagnosis codes similar to `DIAGNOSES.csv`. The `MIMICExtractDataset` also includes a few convenient "wrapper" functions (`parse_labevents` and `parse_chartevents`) that generate events from the `vitals_labs` dataset that would have made sense to appear in CHARTEVENTS and LABEVENTS in the raw .csv files—however as the code comments state, because of the aggregation that happens, these are approximate representations. The idea of these functions is to make it easy to switch some pipelines that were using `MIMIC3Dataset` to use `MIMICExtractDataset` easily or without modification.

One note about your questions about the original MIMIC-III .csv files: while it is not required to have them, it has occurred to me that—since the MIMIC-Extract data shares the same IDs—it would be possible in the future to allow this class to *also* parse data from the original .csv files to add additional `Event` entries—for instance, the MIMIC-Extract data does not include any prescription data, but this could easily be added from PRESCRIPTIONS.csv file for the patients in the selected cohort...just an idea for the future.

We can chat on Zoom if you'd like, I'm available this week after 5pm CDT.

Switching this PR from "draft" as requested, sorry I forgot that yesterday.

ycq091044:
Thanks for the explanation @SphtKr. I have set up a short zoom call at 17:15 this afternoon CDT 

Topic: MIMICExtract Dataset - with SphtKr
Time: May 2, 2023 05:15 PM Central Time (US and Canada)

Join Zoom Meeting
https://illinois.zoom.us/j/87375838639?pwd=K0drejdnR2paOTQzZGNqL1JyTjd4Zz09

Meeting ID: 873 7583 8639
Password: 300439

----
Just send an email to me <chaoqiy2@illinois.edu> if you have time and I can schedule another meeting.

ycq091044 (approved):
Sorry for the delay. The current PR looks good. Thanks very much for your contribution!

SphtKr:
@ycq091044 Thanks! It doesn't appear that I have merge permissions on the repo so please squash-and-merge for me!

PR #140, opened 2023-04-26
fix mask create bug

louis-she:
I was faced a bug that the sum of `dim` is 0 but the elements in dim are not zero. If using `sum` to get mask, this case will cause bugs. The right way is to use `torch.any(x !=0, dim=2)`.

### DIFF
--- a/pyhealth/models/adacare.py
+++ b/pyhealth/models/adacare.py
@@ -460,7 +460,7 @@ class AdaCare(BaseModel):
                 # (patient, event, embedding_dim)
                 x = self.embeddings[feature_key](x)
                 # (patient, event)
-                mask = torch.sum(x, dim=2) != 0
+                maks = torch.any(x !=0, dim=2)
 
             # for case 2: [[code1, code2], [code3, ...], ...]
             elif (dim_ == 3) and (type_ == str):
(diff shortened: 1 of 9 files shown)


ycq091044 (changes requested):
Very hidden bug, though it seems the bug is very hard to reproduce. Thanks! Please change "maks" to "mask".

louis-she:
Sorry for the typo, already fixed.

> though it seems the it is hard to reproduce.

yes it is. I was running a very long parameter tunning process and after 4 or 5 days running I encountered this bug.

ycq091044:
@louis-she Could I ask what is the error message? It seems that even the mask is (hiddenly) calculated wrong, the model can still run (with the wrong mask). Theoretically, both `sum` and `any` function have an extremely low probability to give wrong mask.

louis-she:
Some samples in my sequential dataset have a sequence length of 1, so the shape of the input batch will be `(bs, 1, dim)`.
When a sample's embedding of that batch is summed to a value that is smaller than `1e-5`, the `!= 0` will give a `False` cause of float point precision, then the only element in the "sequence" will be masked. This will lead to undefined behaviors, cause all of the sequences are masked.

Specifically, the error I met is raised at:

```
rx = rnn_utils.pack_padded_sequence(
    rx, lengths, batch_first=True, enforce_sorted=False
)
```
there is a 0 in the `lengths` list, and the `torch` complains that all the elements in `lengths` should be larger than 0. After I changed to my version of the code, everything went well with the same settings, fixed seed.

louis-she:
image

( I was helping others to debug their code so I do not own the source code, can only provide screenshots )

ycq091044:
Makes sense. Thanks, I will merge the PR right now.

>  I was helping others to debug their code so I do not own the source code, can only provide screenshots.

Impressive...

ycq091044 (approved):

PR #141, opened 2023-04-27
Fix GRASP model's bugs (type hints, parameter error, infinite loop er…

ycq091044:
…rors) (#139)

* Fix GRASP model `__init__` parameter type, `block` should be a string

* Fix GRASP model, if use RNNLayer the returned params sequence wrong

RNNLayer returns outputs, last_outputs, here we need the last_outputs when self.backbone is called.

* Fix GRASP model's GRASPLayer's return type, `torch.tensor` not a Tuple

GRASPLayer returns a single tensor: out, not a tuple type

* Fix GRASP model missing `num_iterations += 1`

The missing argument may cause infinite loop

### DIFF
--- a/pyhealth/models/grasp.py
+++ b/pyhealth/models/grasp.py
@@ -79,6 +79,7 @@ def cluster(dataset, num_centers, device):
     codes = compute_codes(dataset, centers)
     num_iterations = 0
     while True:
+        num_iterations += 1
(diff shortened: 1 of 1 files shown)


ycq091044:
@yhzhu99  Thanks for correcting this. I merged it to this branch first for local test (which is a success). I will merge it.

PR #147, opened 2023-05-01
Added new parser functions and prediction tasks for eICU dataset

lycpaul:
I am working on a model that required parsing admissionDx and diagonsisString table in eICU dataset: [paper](https://arxiv.org/pdf/1906.04716.pdf), [source code](https://github.com/Google-Health/records-research/tree/master/graph-convolutional-transformer). As I am using PyHealth's model as the baseline, I would like to include the proper parser and prediction tasks in the preprocessing pipeline.


ycq091044:
Thanks for your pr. I will merge into an intermediate branch for local test.

PR #148, opened 2023-05-02
Addition for eicu

ycq091044:
merge into this branch for local test.

### DIFF
--- a/pyhealth/datasets/eicu.py
+++ b/pyhealth/datasets/eicu.py
@@ -34,7 +34,7 @@ class eICUDataset(BaseEHRDataset):
 
     We further support the following tables:
         - diagnosis: contains ICD diagnoses (ICD9CM and ICD10CM code)
-            for patients
+            and diagnosis information (under attr_dict) for patients
         - treatment: contains treatment information (eICU_TREATMENTSTRING code)
             for patients.
         - medication: contains medication related order entries (eICU_DRUGNAME
@@ -45,8 +45,6 @@ class eICUDataset(BaseEHRDataset):
             conducted for patients.
         - admissionDx:  table contains the primary diagnosis for admission to
             the ICU per the APACHE scoring criteria. (eICU_ADMITDXPATH)
-        - diagnosisString: contains the diagnosis information (eICU_DIAGNOSISSTRING)
-            for patients
 
     Args:
         dataset_name: name of the dataset.
@@ -81,7 +79,7 @@ class eICUDataset(BaseEHRDataset):
         >>> from pyhealth.datasets import eICUDataset
         >>> dataset = eICUDataset(
         ...         root="/srv/local/data/physionet.org/files/eicu-crd/2.0",
-        ...         tables=["diagnosis", "medication", "lab", "treatment", "physicalExam", "admissionDx", "diagnosisString"],
+        ...         tables=["diagnosis", "medication", "lab", "treatment", "physicalExam", "admissionDx"],
         ...     )
         >>> dataset.stat()
         >>> dataset.info()
@@ -245,10 +243,10 @@ class eICUDataset(BaseEHRDataset):
         # read table
         df = pd.read_csv(
             os.path.join(self.root, f"{table}.csv"),
-            dtype={"patientunitstayid": str, "icd9code": str},
+            dtype={"patientunitstayid": str, "icd9code": str, "diagnosisstring": str},
         )
         # drop rows with missing values
-        df = df.dropna(subset=["patientunitstayid", "icd9code"])
+        df = df.dropna(subset=["patientunitstayid", "icd9code", "diagnosisstring"])
         # sort by diagnosisoffset
         df = df.sort_values(["patientunitstayid", "diagnosisoffset"], ascending=True)
         # add the patient id info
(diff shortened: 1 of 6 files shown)


pyhealth/datasets/eicu.py
```diff
@@ -563,11 +567,148 @@ def physicalExam_unit(v_info):
...
+
+            events = []
+            for offset, code in zip(
+                    v_info["diagnosisoffset"], v_info["diagnosisstring"]
+            ):
+                # compute the absolute timestamp
+                timestamp = v_encounter_time + pd.Timedelta(minutes=offset)
+                event = Event(
+                    code=code,
+                    table="diagnosisString",
```
ycq091044: minor: change line653 and use table=table.

pyhealth/datasets/eicu.py
```diff
@@ -563,11 +567,148 @@ def physicalExam_unit(v_info):
...
+        # summarize the results
+        patients = self._add_events_to_patient_dict(patients, group_df)
+        return patients
+
 
 if __name__ == "__main__":
     dataset = eICUDataset(
         root="/srv/local/data/physionet.org/files/eicu-crd/2.0",
-        tables=["diagnosis", "medication", "lab", "treatment", "physicalExam"],
+        tables=["diagnosis", "medication", "lab", "treatment", "physicalExam", "admissionDx", "diagnosisString"],
```
ycq091044: May I know where could I find the diagnosisString.csv table? It does not exist in the eicu-2.0 dataset.

pyhealth/tasks/mortality_prediction.py
```diff
@@ -184,6 +184,71 @@ def mortality_prediction_eicu_fn(patient: Patient):
...
+            continue
+        # TODO: should also exclude visit with age < 18
+        samples.append(
+            {
+                "visit_id": visit.visit_id,
+                "patient_id": patient.patient_id,
+                "conditions": [admissionDx] + [diagnosisString],
+                "procedures": [treatment],
+                "label": mortality_label,
+            }
```
ycq091044, line 258: I would suggest merging this mortality2 func with the original mortality. The final samples can have more keys "conditions", "procedures", "drugs", "admDX+diagString", "treatment", ... Later, user can choose a subset as their feature keys and choose drugs or treatment as their target key.

pyhealth/tasks/readmission_prediction.py
```diff
@@ -187,6 +187,71 @@ def readmission_prediction_eicu_fn(patient: Patient, time_window=5):
...
+        # TODO: should also exclude visit with age < 18
+        samples.append(
+            {
+                "visit_id": visit.visit_id,
+                "patient_id": patient.patient_id,
+                "conditions": [admissionDx] + [diagnosisString],
+                "procedures": [treatment],
+                "label": readmission_label,
+            }
+        )
```
ycq091044, line 262: the same here, please merge two readmission funcs.

ycq091044 (commented):
Thanks for your contribution. @lycpaul Please improve the PR following the comments.

pyhealth/datasets/eicu.py
lycpaul: It is under the [diagnosis.csv](https://eicu-crd.mit.edu/eicutables/diagnosis/). Instead of using the ICD9Code, the paper used the "diagnosisString" column instead.

pyhealth/tasks/mortality_prediction.py
lycpaul: It's feasible, just that it will drop a lot of data as it only keeps data containing both entries: 
```python
if len(conditions) * len(procedures) * len(drugs) == 0:
        continue
```
I am afraid it will break some of the previous pipeline/model performance, so I created another task instead. Another option is to pass a feature key set, like ```key_set = {'condition: ['admissionDx', 'diagnosis], 'procedures' = ['treatment', 'medical']}```, to indicate the features we want to keep. But that would change the unified format of the tasks class.

pyhealth/datasets/eicu.py
ycq091044: I see, could you please merge the diagnosis and diagnosisString. It could be done by:
1. add an extra diagnosisString-value pair in the diagnosis Event https://pyhealth.readthedocs.io/en/latest/api/data/pyhealth.data.Visit.html
2. instead of using "get_code_list" in the task_func, please use "get_event_list" and then get the value from the "diagnosisString" key.

pyhealth/tasks/mortality_prediction.py
ycq091044: You are right. Let us keep two functions then. 

Could you please help add minor descriptions (condition represents which table? procedure represents which table? etc) below each "def readmission_prediction_eicu_fn" to differentiate two functions?

lycpaul:
Thanks for the comments, I have resolved them in (#149)

pyhealth/tasks/mortality_prediction.py
```diff
@@ -184,6 +189,72 @@ def mortality_prediction_eicu_fn(patient: Patient):
...
+
+        # exclude: visits without treatment, admissionDx, diagnosisString
+        if len(admissionDx) * len(diagnosisString) * len(treatment) == 0:
+            continue
+        # TODO: should also exclude visit with age < 18
+        samples.append(
+            {
+                "visit_id": visit.visit_id,
+                "patient_id": patient.patient_id,
+                "conditions": [admissionDx] + [diagnosisString],
```
ycq091044: `admissionDx` is a list and `diagnosisString` is also a list. [admissionDx] + [diagnosisString] will gives [[xxxx], [xxx]].

You can use either `admissionDx + diagnosisString` or `[admissionDx + diagnosisString]` here.

PR #157, opened 2023-05-14
fix very low accuracy metric

louis-she:
The accuracy of multilabel task is ~0.002 no matter how tuned the parameters, comparing to the ~0.91 on the [benchmark](https://pyhealth.readthedocs.io/en/latest/#benchmark-on-healthcare-tasks). The reason for the low accuracy is that input of `sklearn_metrics.accuracy_score` is not flattened. Given a 200 classes prediction, the accuracy counts right when the 200 is exactly matched the ground truth. If we flatten the predictions and targets before calling `sklearn_metrics.accuracy_score`, then the accuracy is about the same as the benchmark.

Also, I have added some logging enhancement in this PR.

### DIFF
--- a/pyhealth/__init__.py
+++ b/pyhealth/__init__.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from pathlib import Path
+import sys
 
 __version__ = "1.1.3"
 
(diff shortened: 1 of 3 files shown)


zzachw:
Looks good to me. Thanks for the PR!

ycq091044 (approved):
LGTM

PR #161, opened 2023-05-31
OMOP measurement continuous values

BPDanek:
Per, http://ohdsi.github.io/CommonDataModel/cdm53.html#MEASUREMENT, we can add an attribute for continuous values. Currently, the pyhealth hosted demo data used for tests does not leverage this field, and so these changes do not affect the unit tests


pyhealth/datasets/omop.py
```diff
@@ -430,7 +454,7 @@ def measurement_unit(p_info):
 
 if __name__ == "__main__":
     dataset = OMOPDataset(
-        root="/srv/local/data/zw12/pyhealth/raw_data/synpuf1k_omop_cdm_5.2.2",
+        root="https://storage.googleapis.com/pyhealth/synpuf1k_omop_cdm_5.2.2/",
```
BPDanek, line 457: Happy to remove this if needed.

pyhealth/datasets/omop.py
```diff
@@ -405,17 +408,38 @@ def measurement_unit(p_info):
...
+                        )
+                    else: 
+                        event = Event(
+                            code=code,
+                            table=table,
+                            vocabulary="MEASUREMENT_CONCEPT_ID",
+                            visit_id=v_id,
+                            patient_id=p_id,
+                            timestamp=strptime(timestamp),
+                        )
```
ycq091044, line 442: We will call the measurements codes using the vocabulary name, so they have to be the same, such as "MEASUREMENTS_CONCEPT_ID". Could we use the following?

```
event = Event(
    code=code,
    table=table,
    vocabulary="MEASUREMENTS_CONCEPT_ID",
    visit_id=v_id,
    patient_id=p_id,
    timestamp=strptime(timestamp),
    measurement=value_numeric if not math.isnan(value_numeric) elif value_categorical else: None
)
```

ycq091044:
BTW, could you please fix the omop doc type in this pr as well (similar to https://github.com/sunlabuiuc/PyHealth/pull/166). I accidentally merged it into the develop branch. But the develop is deprecated.

PR #169, opened 2023-06-06
fix bugs, add new metrics and tasks

v1xerunt:
1. fix mismatch between actual output and document descriptions in adacare, concare, agent and stagenet
2. add new metrics: outcome-specific mean absolute error, early prediction score from paper "A Comprehensive Benchmark for COVID-19 Predictive Modeling Using Electronic Health Records in Intensive Care"
3. add regression related metrics
4. add new task: outcome-specific length of stay prediction
5. change base_model to support new tasks


pyhealth/models/base_model.py
```diff
@@ -242,32 +245,44 @@ def get_loss_function(self) -> Callable:
...
-                - a list of str for binary and multiclass classificationa
+                - a list of str for binary and multiclass classification
                 - a list of list of str for multilabel classification
             label_tokenizer: the label tokenizer.
 
         Returns:
             labels: the processed labels.
         """
-        if self.mode in ["binary"]:
+        if self.mode in ["binary"] or self.mode in ["early_mortality"]:
```
ycq091044, line 285: self.mode in ["binary", "early_mortality"]

pyhealth/models/base_model.py
```diff
@@ -281,36 +296,95 @@ def prepare_labels(
...
+            - multi_target: a list of tensors with the same format as above
 
         Args:
             logits: the predicted logit tensor.
 
         Returns:
             y_prob: the predicted probability tensor.
         """
-        if self.mode in ["binary"]:
+        if self.mode in ["binary"] or self.mode in ["early_mortality"]:
```
ycq091044, line 330: same as above

PR #175, opened 2023-06-22
Add models for image classification task

wuzijian1997:
1. Add torchvision_classif.py to models. This file integrated ResNet, DenseNet, Vision Transformer, Swin Transformer, and Swin Transformer V2 based on the Torchvision library.
2. Delete the previous resnet.py


zzachw (approved):
LGTM. Thanks Zijian for the PR!

PR #177, opened 2023-07-02
add cardiology dataset, task, example (#176)

ycq091044:
transfer Yanjing's PR from #176


ycq091044:
I improved your code a bit and add readthedocs documentations for the new datasets and tasks. Thanks @YanjingLiLi, very impressive work.

PR #193, opened 2023-07-18
Add MedCLIP Model for Zero-shot Inference

wuzijian1997:
Add a new model medclip_zeroshot.py

Note that the original MedCLIP repository seems to have bugs. I mainly modify the "MedCLIPFeatureExtractor" class. The source code lacks some image transformation functions, so passing the image transformation will get errors. I kindly remind you to fix this issue for the original repository.


wuzijian1997:
I add an example of the MedCLIP zero-shot classification. Please check it.

PR #198, opened 2023-07-30
FavMac (ICML 2023)

zlin7:
Implementation of "Fast Online Value-Maximizing Prediction Sets with Conformal Cost Control (FavMac)"


zlin7:
One can test FavMac by running `__init__.py`

PR #225, opened 2023-10-02
Chefer Transformer Gradient-Weighed Self Attention Interpretability

jhnwu3:
I created an interpret module that is unfortunately half-baked as I haven't had time to test across multiple datasets. I have modified the models/transformer.py to allow for easy computation of gradients and a CheferRelevance class.


jhnwu3:
Cleaned up any bugs (specifically missing a + operator in relevance aggregation). Also deleted a bunch of unnecessary files that I don't have time to implement right now, but I feel the Chefer Relevance base class is still a needed step up in being able to interpret PyHealth's transformers, at least, at the token level. The demo should be in the examples/ directory.

The visualization stuff will need to be implemented later. Thanks for the hackathon guys, I normally wouldn'tve taken the time to actually learn someone else's codebase, but I at least know where to look if I need to quickly just do a health ML project now.

PR #226, opened 2023-10-02
add graph model

Namkyeong:
I added Neighborhood Sampler and graph_torchvision_model.py for graph learning


zzachw:
The model looks good to me. I will combine the dataset part with  #172 and merge later.

PR #228, opened 2023-10-07
Add KDD'23 paper "MedLink: De-Identified Patient Health Record Linkage"

zzachw:
The following changes are made:
1. Dataset: `MIMIC3Dataset` is slightly modified to include additional patient identifiers
2. Task: Add a new task file `patient_linkage.py`
3. Model:
3.1 Add a new model folder `medlink`, which includes the MedLink model as well as some supporting functions
3.2 `BaseModel` is slightly modified to accommodate the retrieval task
4. Training: `Trainer` is slightly modified to accommodate the retrieval task
5. Inference:
5.1 Add a new metric file `ranking.py`
5.2 MedLink uses its own evaluate function for this particular retrieval task


ycq091044:
@zzachw Could you add a notebook for it?

PR #239, opened 2023-10-21
231021 chefer transformer

ycq091044:
Directly merged from John's PR #225


ycq091044:
Hi @jhnwu3, I can run your demo jupyter successfully. However, what is the relevance score here? Does higher score mean the feature is more important?

image

PR #240, opened 2023-10-21
231031 medlink

ycq091044:
merged from #228


pyhealth/models/base_model.py
```diff
@@ -13,7 +13,7 @@
 
 
 # TODO: add support for regression
-VALID_MODE = ["binary", "multiclass", "multilabel"]
+VALID_MODE = [None, "binary", "multiclass", "multilabel"]
```
ycq091044: Could we replace "None" with "Ranking" or something else?

pyhealth/metrics/ranking.py
```diff
@@ -0,0 +1,38 @@
+from typing import List, Dict
+
+
+def ranking_metrics_fn(qrels: Dict[str, Dict[str, int]],
```
ycq091044, line 4: Could you add an example for this ranking function? How to use it, what is query, result, and kvalues?

pyhealth/trainer.py
```diff
@@ -304,12 +306,22 @@ def evaluate(self, dataloader) -> Dict[str, float]:
         Returns:
             scores: a dictionary of scores.
         """
-        y_true_all, y_prob_all, loss_mean = self.inference(dataloader)
-
-        mode = self.model.mode
-        metrics_fn = get_metrics_fn(mode)
-        scores = metrics_fn(y_true_all, y_prob_all, metrics=self.metrics)
-        scores["loss"] = loss_mean
+        if self.model.mode is not None:
```
ycq091044, line 309: change into your new mode, but do not use  "None"

ycq091044 (commented):
LGTM, please address the minor comments and add docstring if possible.

pyhealth/models/base_model.py
zzachw: The evaluation pipeline for ranking is not compatible with Pyhealth's Trainer. That's why I add a None option here. I will add more instructions.

zzachw:
Resolved. Can you take a look? @ycq091044

PR #281, opened 2024-04-17
CS598-DLH Bonus Points PR for Text2MolDataset with text2mol_dataset.py

darinz:
Submitting PR for New dataset (10 points) bonus points for CS598-DLH class. The new dataset is Text2MolDataset.

### DIFF
--- a/pyhealth/datasets/text2mol_dataset.py
+++ b/pyhealth/datasets/text2mol_dataset.py
@@ -1,0 +1,284 @@
+"""
+File: text2mol_dataset.ipynb
+
+PyHealth Text2Mol Dataset
+"""
+
+import numpy as np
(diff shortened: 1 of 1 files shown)


ycq091044 (commented):
Thanks for the PR @darinz.

I have tested the PR, which runs successfully. The PR alone is great! You do not have to modify anything and we will give bonus points to you.

Generally, we would ask all new dataset classes to inherit from pyhealth based dataset class. However, we have not finished the multi-modality base dataset yet, so the PR will be left here and could be merged into master branch once we have multi-modality supports for it.

ycq091044:
Merged into 240506-sp24-course-pr for now. Will need to refactor based on pyhealth base dataset @zzachw

PR #292, opened 2024-05-08
added condition recommendation task

san2ds:


ycq091044:
Hello @san2ds , thanks for your PR. Could I ask the motivation of the task?

san2ds:
Hi,

We have used this task to create dataset using feature keys - Condition
history, Procedure and prescription from past visit history of a patient.
We have used that in our Multilabel Prediction model (Model 1) of diagnosis
code of next visit using Pyhealth.

Please find our github link of final project submission:

https://github.com/jt-alvarez/CS598-DLH-Project/blob/main/DL4H_Team_152_combine.ipynb

We have used that to reproduce result of paper TransformEHR.
https://www.nature.com/articles/s41467-023-43715-z

In addition to that task we have created 4 more tasks also. We have not
created a PR on that yet. Please let me know if you think we should add PR
for those other tasks also.

Task Function 1 - Task Function for diagnosis recommendation of future
visits using past visit's Condition history, Procedure and prescription
history.
Task Function 2 - Task Function for diagnosis recommendation using only
history of diagnosis from past visits.
Task Function 3 - Task Function for diagnosis recommendation related to
Pancreatic Cancer using only history of diagnosis from past visits.
Task Function 4 - Task Function for diagnosis recommendation related to
Pancreatic Cancer using history of Diagnosis, Procedure and Drugs
Prescribed from past visits.
Task Function 5 - Task Function for diagnosis recommendation related to
Pancreatic Cancer using only history of diagnosis from past visits. This
will add data samples with and without Pancreatic Cancer Diagnosis. So that
model performance will be done more accurately.



Thank you!

Santu Banerjee

ycq091044:
Thanks for your comments.

PR #336, opened 2025-04-15
Refactor MedicalTranscriptionsDataset

zzachw:
- Updated MedicalTranscriptionsDataset to the new API
- Adjusted TableConfig to allow patient_id to be optional, defaulting to row index if not provided.
- Fixed bug in MultiClassLabelProcessor


jhnwu3 (approved):
Looks good to me!

PR #337, opened 2025-04-15
Refactor COVID19CXRDataset

zzachw:
- Updated COVID19CXRDataset to the new API
- Updated TorchvisionModel to the new API
- Updated  ImageProcessor docstring
- Added progress indication in SampleDataset during sample processing.

Note:
1. The `ImageProcessor` currently loads all images into memory before the training iteration. We should add a online loading option later such that user can load images from disk during training.
2. We should allow `ImageProcessor` to automatically mean and std for normalization using the `fit()` function later.

I will work on 1 & 2 later in a separate PR.

### DIFF
--- a/pyhealth/datasets/configs/covid19_cxr.yaml
+++ b/pyhealth/datasets/configs/covid19_cxr.yaml
@@ -1,0 +1,10 @@
+version: "5.0"
+tables:
+  covid19_cxr:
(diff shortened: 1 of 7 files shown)


jhnwu3 (approved):
Looks good to me. Will see if I can't make the documentation surrounding the metatable idea more clear for new contributors.

PR #342, opened 2025-04-19
Add HurtfulWordsDataset and MaskedBiasScoreTask for masked language model evaluation

zelal-Eizaldeen:
- I am Zilal Eiz Al Din - with NetId: zelalae2@illinois.edu
- I have added a Contribution to Datasets by adding the HurtfulWordsDataset Class, and a Contribution to the Tasks by adding the MaskedBiasScoreTask Class with unittests (test_hurtfulwords_cleaning_data.py and test_hurtfulwords_logP_task.py) and an example (masked_gender_bias_score.py)
- I added these contributions based on the paper called 'Hurtful Words: Quantifying Biases in Clinical Contextual Word
Embeddings' to have a dataset with MASK gendered words (he, she, male, female.. ) and to calculate the log probability of the masked word as a task.
- Includes unit tests and example scripts.
- Rebased with latest sunlabuiuc/main.
- Allow maintainers to edit this PR for minor changes.
- To test the code, under unittests folder, there is two files called test_hurtfulwords_cleaning_data.py and test_hurtfulwords_logP_task.py), you can run them.
- To test the code, under examples folder, there is a file called masked_gender_bias_score.py.

### DIFF
--- a/examples/masked_gender_bias_score.py
+++ b/examples/masked_gender_bias_score.py
@@ -1,0 +1,69 @@
+# ------------------------------------------------------------------------------
+# Example Script: HurtfulWordsDataset + MaskedBiasScoreTask
+#
+# Name:
+#   Zilal Eiz Al Din
+#
+# NetID:
+#   zelalae2
+#
+# Purpose:
+#   This script demonstrates how to:
+#   - Load the HurtfulWordsDataset (MIMIC-III noteevents) 
+#   - Apply the MaskedBiasScoreTask to compute gender bias scores
+#   - Iterate through the generated samples
+#   - Print patient information and bias-related log-probabilities
+#
+# Paper Reference:
+#   Hurtful Words: Quantifying Biases in Clinical Contextual Word Embeddings
+#   https://arxiv.org/abs/2003.11515
+#
+# Usage:
+#   This is a simple script for manual inspection and demonstration.
+#   To run:
+#       $ python example_hurtfulwords_task.py
+#
+# Note:
+#   This script assumes you have access to MIMIC-III CSVs at the specified root path.
+# ------------------------------------------------------------------------------
+from pyhealth.datasets import HurtfulWordsDataset
+from pyhealth.tasks import MaskedBiasScoreTask
+
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


zelal-Eizaldeen:
Implements HurtfulWordsDataset for masked clinical note preprocessing.

Implements MaskedBiasScoreTask for bias score computation.

jhnwu3:
Hey Zilal and Payel, thanks for the PR!

Some comments, so don't be afraid to correct me if I'm misunderstanding the contribution.

- The dataset is purely using MIMIC3 right with additional annotations? 
- It seems the HurtfulWords is a "task" based on MIMIC3 in the sense that you guys apply an additional processing step on top of MIMIC3. So, maybe, if you can fit the processing steps into a task class, that makes more sense in my eyes.
- The score task is interesting, but I'm not clear why someone would want to fit the scores of the list of samples of another model? Can ya'll elaborate? It sounds more like a **model** to me rather than a task.

zelal-Eizaldeen:
Thank you for your feedback.

In terms of the dataset, in the paper, we needed to process MIMIC3
noteevents to convert the column TEXT to have MASK label for gender.
Then in the task class, we calculated the log probability of masked words.
I am not sure what does a dataset mean to be as a contribution?

what do you think to do that better?


Kind regards,
Zilal

zelal-Eizaldeen:
In terms of Log probability, in deed I used BERT 
model to calculate the MASK words probability.
All in all, could u plz confirm that I need to move the Hurtful dataset into tasks folder? and the MaskedBoasscore task into models folder?

jhnwu3:
Hey Zilal, yes, I think that would make more sense. 

Since in the task class, you can do the masking of the existing text samples and have it returned. 

The reason why I think computing gender logprob differences is a separate model is because in some sense you're computing a bias score with an existing model rather than it being a set of annotated scores or measurements that would typically fit in the task class.

zelal-Eizaldeen:
Added HurtfulWordsPreprocessingTask and FillInTheBlank Model with examples and unittests.

jhnwu3 (commented):
Excellent PR. Highlighted.

PR #335, opened 2025-04-14
MIMIC4 Refactor to Encompass Vision, Language, and EHR

jhnwu3:
This pull request introduces several enhancements and new features to the `pyhealth` library, particularly focusing on the `BaseDataset` class, configuration files, and task definitions. The most important changes include adding a development mode to limit the number of patients, updating configuration files for new datasets, and improving the structure of task definitions.

Enhancements to `BaseDataset` class:

* [`pyhealth/datasets/base_dataset.py`](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR26): Added a `dev` parameter to the `BaseDataset` class to enable development mode, which limits the dataset to 1000 patients. This includes updates to the `__init__` method, `collected_global_event_df`, and logging statements to reflect the development mode status. [[1]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR26) [[2]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR35) [[3]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR44-R53) [[4]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dL65-R85) [[5]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR191) [[6]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR231) [[7]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dL234-R279)

Updates to configuration files:

* [`pyhealth/datasets/configs/mimic4_cxr.yaml`](diffhunk://#diff-a627cb6817a87419df3248f1975e70ef23c954918b6bd67df950b383fcc8396dR1-R91): Added new configuration entries for X-ray metadata, CheXpert labels, NegBio labels, test set labels, and image paths.
[...]

PR #350, opened 2025-04-25
Add EHRShot dataset and benchmarking tasks

zzachw:
- Introduced EHRShotDataset class for managing EHRShot data.
- Added configuration YAML file for EHRShot dataset
- Created BenchmarkEHRShot task for evaluating various clinical prediction tasks
- Developed example notebook for benchmarking EHRShot with XGBoost
- Updated label processors to handle label index to avoid the case where label 1 is mapped to index 0

### DIFF
--- a/pyhealth/tasks/benchmark_ehrshot.py
+++ b/pyhealth/tasks/benchmark_ehrshot.py
@@ -1,0 +1,89 @@
+from typing import Any, Dict, List, Optional
+
+import polars as pl
+
+from .base_task import BaseTask
+
+
+class BenchmarkEHRShot(BaseTask):
+    """Benchmark predictive tasks using EHRShot."""
+
+    tasks = {
+        "operational_outcomes": [
+            "guo_los",
+            "guo_readmission",
+            "guo_icu"
+        ],
+        "lab_values": [
+            "lab_thrombocytopenia",
+            "lab_hyperkalemia",
+            "lab_hypoglycemia",
+            "lab_hyponatremia",
+            "lab_anemia"
+        ],
+        "new_diagnoses": [
+            "new_hypertension",
+            "new_hyperlipidemia",
+            "new_pancan",
+            "new_celiac",
+            "new_lupus",
+            "new_acutemi"
+        ],
+        "chexpert": [
+            "chexpert"
+        ]
+    }
+
+    def __init__(self, task: str, omop_tables: Optional[List[str]] = None) -> None:
+        """Initialize the BenchmarkEHRShot task.
+
+        Args:
+            task (str): The specific task to benchmark.
+            omop_tables (Optional[List[str]]): List of OMOP tables to filter input events.
+        """
+        self.task = task
+        self.omop_tables = omop_tables
+        self.task_name = f"BenchmarkEHRShot/{task}"
+        self.input_schema = {"feature": "sequence"}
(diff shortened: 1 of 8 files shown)


pyhealth/tasks/benchmark_ehrshot.py
jhnwu3, line 1: I see, so for something like EHR shot, you've created a combo task? That people can characterize? 

I should ask, for MIMIC, does it make for each dataset to have their own task file or similar? I suppose the class is supposed to be flexible for people to do whatever.

zzachw: This BenchmarkEHRShot allows some flexibility underlying the same task definition. It allows the user to select which task to perform and which feature to use. But if the user wants to utilize features other than the code, they need to write their own task class (and processors).

The EHRShot dataset is different from the MIMIC datasets in the sense that it provides task definition (i.e., those four categories of tasks). On the other hand, MIMIC doesn't provide any official task dentition and it is up to the user to decide.

zzachw: For MIMIC, we can also have __init__ function that accepts user args which gives the user some flexibility in customizing tasks using existing templates. For example, we can allow them to select tables / prediction window length, etc.

But generally speaking, we should still let the user know that the classes insides tasks module is more like examples. And the users are encouraged to create their own task classes.

jhnwu3 (commented):
EHRShot runs, but testing mimic3 mortality prediction fails (?) 

Any ideas why that might happen, it's probably not related to the thread creator, but I'm not sure.

```polars.exceptions.InvalidOperationError: conversion from `str` to `datetime[μs]` failed in column 'chartdate' for 1 out of 1 values: ["2024-11-2913:48:49"]

You might want to try:
- setting `strict=False` to set values that cannot be converted to `null`
- using `str.strptime`, `str.to_date`, or `str.to_datetime` and providing a format string```

zzachw:
> EHRShot runs, but testing mimic3 mortality prediction fails (?)
> 
> Any ideas why that might happen, it's probably not related to the thread creator, but I'm not sure.
> 
> ```polars.exceptions.InvalidOperationError: conversion from `str` to `datetime[μs]` failed in column 'chartdate' for 1 out of 1 values: ["2024-11-2913:48:49"]
> 
> You might want to try:
> 
> * setting `strict=False` to set values that cannot be converted to `null`
> * using `str.strptime`, `str.to_date`, or `str.to_datetime` and providing a format string```

The problem comes from this PR #335.

For MIMIC-III `noteevents` table, We set `timestamp` to be a combination of `chartdate` (e.g., `2024-11-29`) and `charttime` (e.g., `13:48:49`).

https://github.com/sunlabuiuc/PyHealth/blob/1ac3ac8d479a444b950d33d4d6e5167e13e40ae0/pyhealth/datasets/configs/mimic3.yaml#L119-L121 

In `BaseDataset.load_table()`, we directly concat `chartdate` and `charttime` (e.g., `2024-11-2913:48:49`).

https://github.com/sunlabuiuc/PyHealth/blob/1ac3ac8d479a444b950d33d4d6e5167e13e40ae0/pyhealth/datasets/base_dataset.py#L157

Since it is not related to this EHRSHOT PR, I will create another PR to solve this.

PR #354, opened 2025-04-26
MIMIC4CXRDataset In-memory Metadata & Overwritable default_tables

Brandon-Galloway:
NetId: bwg2
Type of contribution: Other/feature enhancement
Description: 
  - Updated mimic3/4 datasets to support overwriting default tables when utilizing a custom config
  - Modified MIMIC4CXRDataset's metadata process to eliminate the need for disk i/o when creating derived metadata columns.
    - Original file is processed on init with polars lazy-values configured for filepath and studytime, returned through an implementation of load_table 
Files to look at: mimi3.py, mimic4.py

### DIFF
--- a/pyhealth/datasets/mimic3.py
+++ b/pyhealth/datasets/mimic3.py
@@ -42,8 +42,8 @@ class MIMIC3Dataset(BaseDataset):
         if config_path is None:
             logger.info("No config path provided, using default config")
             config_path = Path(__file__).parent / "configs" / "mimic3.yaml"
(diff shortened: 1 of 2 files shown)


Brandon-Galloway:
Small Update: I've opened a follow-up PR to extend this work — [PR #410](https://github.com/sunlabuiuc/PyHealth/pull/410).

It builds on the in-memory metadata and configurable default_tables features by adding support for extracting bias-related metadata from MIMIC-CXR for downstream fairness and bias mitigation workflows

Thank you!

PR #358, opened 2025-04-27
Create additional_data_accuracy_test.py.

isaacdchan:
Creates a task which demonstrates that a model trained on dataset A1 and A2 is less accurate than a model trained on just dataset A1

Based on the findings from "When More is Less: Incorporating Additional Datasets Can Hurt Performance By Introducing Spurious Correlations"

Part of final project contribution for CS 598 for drcook2 and isaacc5

### DIFF
--- a/pyhealth/tasks/additional_data_accuracy_test.py
+++ b/pyhealth/tasks/additional_data_accuracy_test.py
@@ -1,0 +1,85 @@
+import os
+import gc
+import random
+import numpy as np
+import torch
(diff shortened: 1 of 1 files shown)


jhnwu3 (commented):
Hey guys, if possible could you guys add some documentation one what is it you're doing here?

isaacdchan:
Hi @jhnwu3 just added some documentation. Thanks!

Issue #366, opened 2025-04-30
Allowing loading from both .csv and .csv.gz


zzachw:
For datasets like MIMIC-III and MIMIC-IV, we should allow loading from both .csv or .csv.gz depending what we find in the root folder.

E.g., https://github.com/sunlabuiuc/PyHealth/blob/1ac3ac8d479a444b950d33d4d6e5167e13e40ae0/pyhealth/datasets/configs/mimic3.yaml#L4

PR #368, opened 2025-04-30
Fix issues with `BaseDataset.load_table()`

zzachw:
Follow up with PR https://github.com/sunlabuiuc/PyHealth/pull/350#pullrequestreview-2808710843

1. Introduced `scan_csv_gz_or_csv` function to handle both CSV and CSV.gz files.
2. Added a preprocessing function call to `BaseDataset.preprocess_{table_name}` that allows table specific customized processing
3. Raised a warning if we have to lower case the col names

- Solves issue #366 caused in PR #335 
- (Partially) solves issue #341: notify the user that we have to lower case the col names by raising a warning

### DIFF
--- a/pyhealth/datasets/base_dataset.py
+++ b/pyhealth/datasets/base_dataset.py
@@ -1,14 +1,13 @@
 import logging
 import os
-import requests
-
 from abc import ABC
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
-from typing import Iterator, List, Optional, Union
-from urllib.parse import urlparse
+from typing import Iterator, List, Optional
+from urllib.parse import urlparse, urlunparse
 
 import polars as pl
+import requests
 from tqdm import tqdm
 
 from ..data import Patient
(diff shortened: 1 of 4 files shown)


zzachw:
Note: 

This PR makes the dataset no longer support reading from URL.

I think we should not do that in the first place as it incurs costs on GCP.

The user should download the files first.

https://github.com/sunlabuiuc/PyHealth/blob/f6e277afe6ac7865b11052ffd88c3dfdae2be2d4/pyhealth/datasets/base_dataset.py#L150

pyhealth/datasets/base_dataset.py
```diff
@@ -15,6 +16,34 @@
...
+        pl.LazyFrame: The LazyFrame for the CSV.gz or CSV file.
+    """
+    if isinstance(csv_path, str):
+        csv_path = Path(csv_path)
+    if csv_path.exists():
+        df = pl.scan_csv(csv_path, infer_schema=False)
+    # Try with .csv extension if .csv.gz fails
+    elif "".join(csv_path.suffixes[-2:]) == ".csv.gz" and csv_path.with_suffix("").exists():
+        csv_path = csv_path.with_suffix("")
+        logger.info(f"CSV.GZ file not found, using CSV file instead: {csv_path}")
```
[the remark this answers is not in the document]
zzachw: Fixed in commit d861824694a1ac1b1600163e72345ec535517f89

pyhealth/datasets/base_dataset.py
```diff
@@ -16,22 +19,62 @@
...
+        return response.status_code == 200
+    except requests.RequestException:
+        return False
+
+
 def scan_csv_gz_or_csv(csv_path: Union[str, Path]) -> pl.LazyFrame:
     """
     Scans a CSV.gz or CSV file and returns a LazyFrame.
-    It will first try to scan the given csv_path, if it fails and
-    the file has a .csv.gz extension, it will try to scan the file
```
zzachw: I think we should keep this in the docstring?

jhnwu3: Oh, I didn't even notice that. Was so focused on trying to quickly iterate on getting it to to work with Urls, I can add that back later or if you want to do that right now. But, we should make a note of the logic about url calls.

zzachw: Sure let me handle it

jhnwu3 (approved):
Will need a separate PR later to figure out a robust datettime read function since its pretty clear that we these formats can be very weird.

But everything else works as discussed with the load_table() function so far.

zzachw:
> Will need a separate PR later to figure out a robust datettime read function since its pretty clear that we these formats can be very weird.
> 
> But everything else works as discussed with the load_table() function so far.

Implementing a robust datetime parsing function for various formats might be challenging due to potential ambiguities. However, I previously added a configuration option that allows users to specify the datetime format explicitly.

https://github.com/sunlabuiuc/PyHealth/blob/abc5723013df35327844a40984034b7a2661e3f6/pyhealth/datasets/configs/config.py#L56

PR #391, opened 2025-05-03
fix bug with mortality label always being zero if hospital_expire_flag is a string

PranavHerur:
Pranav Herur
pherur2
CaliForest: Calibrated Random Forest for Health Data
[CaliForest Paper](https://pmc.ncbi.nlm.nih.gov/articles/PMC8299436/)

fix bug with mortality label always being zero if `hospital_expire_flag` is a string value

in my testing with the examples `hospital_expire_flag` came in as "0" or "1" and would always get set to 0
meaning every label would become 0


jhnwu3 (approved):
Thanks for the catch! I just ran into this issue with the synthetic dataset we have for some reason.

PR #343, opened 2025-04-19
NIH Chest X-Ray Dataset with Unit Tests & Classification Task [DL4H; SP25]

ethanrasmussen:
### **Pull Request Details:**
**Authored by:** Ethan Rasmussen (ethanmr3@illinois.edu)
**Paper Title & Link:** n/a (_this is a separate, but similar, contribution to my paper/project_)
**Contributed Dataset Details:**
X. Wang, Y. Peng, L. Lu, Z. Lu, M. Bagheri and R. M. Summers, "ChestX-Ray8: Hospital-Scale Chest X-Ray Database and Benchmarks on Weakly-Supervised Classification and Localization of Common Thorax Diseases," 2017 IEEE Conference on Computer Vision and Pattern Recognition (CVPR), Honolulu, HI, USA, 2017, pp. 3462-3471, doi: 10.1109/CVPR.2017.369.
**Primary Dataset Access Link:** [View on Kaggle](https://www.kaggle.com/datasets/nih-chest-xrays/data)

### Contribution Overview:
 - Adds the NIH-CXR dataset class **"NIHChestXrayDataset"** with the file `nih_cxr.py`
 - Adds a unit test class with multiple tests in the **"TestNIHCxrDataset"** class, under the file `test_nih_cxr.py`
 - Adds a basic chest x-ray classification in the **"ChestXrayClassificationTask"** class, under the file `nih_cxr_classification.py`
 - The COVID-19 CXR code currently in PyHealth was used as a reference when creating this contribution

### NIH-CXR Dataset (`nih_cxr.py`):
 - Downloads & extracts the dataset from Kaggle
 - Scans all `images_*/images/` subdirectories to build a map from filename to full file path, and splits on train/test values as instructed
 - Populates `self.patients`, which contains a dictionary mapping image filepath to label
 - Labels are defined as 0 for "No Finding" and 1 for any finding
 - Includes a removable smoke test in its `__main__` block to test functionality

### Unit Tests (`test_nih_cxr.py`):
 - Sets up dummy data in a temporary directory which is torn down after testing
 - Verifies dummy split has proper amount of samples for testing
 - Ensures sample can be properly loaded as instance of `PIL.Image.Image`
 - Ensures that `IndexError` is raised when expected (preventing undefined behavior from out-of-bounds)
 - Tests that `stat()` output is accurate

### CXR Classification Task using NIH Dataset (`nih_cxr_classification.py`):
[...]

PR #392, opened 2025-05-03
ChestX-ray14 Dataset and Binary Classification Task

EricSchrock:
**Author**: Eric Schrock (ejs9@illinois.edu)
**Dataset**: https://nihcc.app.box.com/v/ChestXray-NIHCC/folder/36938765345
**Dataset paper**: https://arxiv.org/abs/1705.02315

## Overview

This PR adds dataset and binary classification task classes for the ChestX-ray14 dataset.

## Dataset (`chestxray14.py` and `chestxray14.yaml`)

1. Optionally downloads the dataset from Box (optionally, just a partial download).
2. Indexes the image paths to their metadata.
3. Initializes the `BaseDataset` class to better integrate with the rest of `PyHealth`.
4. Provides a basic `__main__` block for manual testing.

## Task (`chestxray14_binary_classification.py`)

1. Creates labeled samples for any of the 14 diseases in the dataset.

## Unit Tests (`test_chestxray14.py`)

1. Fabricates test data to avoid image downloads.
2. Tests the public `ChestXray14Dataset` methods.
3. Tests two sample datasets generated with `ChestXray14BinaryClassification`.
    1. One for hernias
    2. The second for Cardiomegaly

## Review Notes

1. Check if my pathing for imports and files (config and data) is correct. [This](https://github.com/EricSchrock/cxr-ml-gzsl/blob/main/PyHealth.ipynb) Jupyter notebook is what I used to setup my environment and test. I'm not sure how well that lines up with the official development and test environment.
2. I couldn't figure out a way to download the image metadata CSV file from Box (downloads a file preview instead) so I created a publicly available mirror of that file on Google Drive. This is clearly not ideal.
    1. Box: https://nihcc.app.box.com/v/ChestXray-NIHCC/file/219760887468
    2. Google Drive (mirror): https://drive.google.com/file/d/1mkOZNfYt-Px52b8CJZJANNbM3ULUVO3f/view?usp=sharing
4. #343 incorporates the same dataset. Here are some key differences.
    1. Their Kaggle download is probably cleaner than my Box / Google Drive download.
    3. Their `NIHChestXrayDataset` class does not initialize the `BaseDataset` class.
    5. They do not provide a table config file.
    6. Their dataset and task combine all fourteen diseases into one "finding" vs "no finding" label.

### DIFF
--- a/pyhealth/datasets/chestxray14.py
+++ b/pyhealth/datasets/chestxray14.py
@@ -1,3 +1,4 @@
+import argparse
 import hashlib
 import logging
 import os
@@ -438,12 +439,15 @@ class ChestXray14Dataset(BaseDataset):
                 },
             })
 
-def main() -> None:
+if __name__ == "__main__":
     logger.setLevel(logging.INFO)
-    dataset = ChestXray14Dataset(partial=True)
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--no-download", action="store_true")
+    args = parser.parse_args()
+
+    dataset = ChestXray14Dataset(download=(not args.no_download), partial=True)
+
     dataset.stat()
     dataset.info()
     print(dataset[0])
-
-if __name__ == "__main__":
-    main()
--- a/pyhealth/datasets/chestxray14.py
+++ b/pyhealth/datasets/chestxray14.py
@@ -112,7 +112,7 @@ class ChestXray14Dataset(BaseDataset):
         dataset_name (str): Name of the dataset.
         paper_url (str): URL of the original paper introducing the dataset.
         dataset_url (str): URL to download the dataset from the NIH repository.
-        path (Path): Filesystem path to the dataset root directory.
+        root (str): Filesystem path to the dataset root directory.
         download (bool): Whether to download the dataset if not already present.
         partial (bool): Whether to download a smaller subset of the dataset.
         transform (Compose): Transformations applied to each image sample.
@@ -134,7 +134,7 @@ class ChestXray14Dataset(BaseDataset):
         >>> from pathlib import Path
         >>> from torchvision.transforms import Compose, Resize, ToTensor
         >>> transform = Compose([Resize((224, 224)), ToTensor()])
-        >>> dataset = ChestXray14Dataset(path=Path("./data"), download=True, transform=transform)
+        >>> dataset = ChestXray14Dataset(root="./data", download=True, transform=transform)
         >>> print(len(dataset))
         >>> image, metadata = dataset[0]
     """
@@ -142,7 +142,7 @@ class ChestXray14Dataset(BaseDataset):
                  dataset_name: str = "ChestX-ray14",
                  paper_url: str = "https://arxiv.org/abs/1705.02315",
                  dataset_url: str = "https://nihcc.app.box.com/v/ChestXray-NIHCC",
-                 path: Path = Path(""),
+                 root: str = "",
                  download: bool = True,
                  partial: bool = False,
                  transform: Compose = None) -> None:
@@ -152,7 +152,7 @@ class ChestXray14Dataset(BaseDataset):
             dataset_name (str): Name of the dataset. Defaults to "ChestX-ray14".
             paper_url (str): URL to the dataset's reference paper. Defaults to the original ChestX-ray14 paper.
             dataset_url (str): URL to download the dataset. Defaults to the NIHCC Box link.
-            path (Path): Local path to store or load the dataset. Defaults to the current directory.
+            root (str): Local path to store or load the dataset. Defaults to the current directory.
             download (bool): Whether to download the dataset or use an existing copy. Defaults to True.
             partial (bool): Whether to download only a subset of the dataset. Defaults to False.
             transform (Compose): Optional torchvision transform pipeline to apply to the images. Defaults to None.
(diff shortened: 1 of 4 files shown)


pyhealth/datasets/chestxray14.py
jhnwu3, line 1: Some thoughts:

If possible, can we format it like the NIH CXR dataset? This is on us for not having the correct documentation set in place (it's on my massive todo list lol), but tldr;

https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/datasets/covid19_cxr.py

Given a directory full of images and thus image paths, we should construct a metadata.csv file (or if there already exists one, update it), such that it has each piece of information with the image pathing as an Event. 

One reason we want to do this is because everything in PyHealth is represented as a list or set of events. 

Here's a tutorial on the events, but in principle, once you've constructed a metadata.csv, the base init process should work accordingly.
https://colab.research.google.com/drive/1y9PawgSbyMbSSMw1dpfwtooH7qzOEYdN?usp=sharing

pyhealth/tasks/chestxray14_binary_classification.py
jhnwu3, line 1: Looks good to me so far.

pyhealth/tasks/chestxray14_multilabel_classification.py
jhnwu3, line 1: Tbh, the only thing that'll change here is that by default everything will be stored as a table of events already, so this will be simplified.

tests/core/test_chestxray14.py
jhnwu3, line 1: Same comment as the other above.

pyhealth/tasks/chestxray14_multilabel_classification.py
```diff
@@ -0,0 +1,65 @@
...
+            ValueError: If the number of chestxray14 events is not exactly one.
+        """
+        events: List[Event] = patient.get_events(event_type="chestxray14")
+        if len(events) != 1:
+            msg = f"Expected just 1 event but got {len(events)}!"
+            logger.error(msg)
+            raise ValueError(msg)
+
+        from ..datasets.chestxray14 import ChestXray14Dataset # Avoid circular import
+        return [{"image": events[0]["path"], "path": events[0]["path"], "labels": [disease for disease in ChestXray14Dataset.classes if int(events[0][disease])]}]
```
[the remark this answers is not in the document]
EricSchrock: @jhnwu3 I believe this Copilot comment is only partially correct. My docstring was wrong (fixed), but I believe my code for the `labels` key is correct. Per your comment in Discord, I did refer to `covid19_cxr_classification.py` when I originally created `chestxray14_multilabel_classification.py`. However, the former is a "multiclass" task, and the latter is a "multilabel" task.

`MultiLabelProcessor` expects `labels` to be a list of label indexes or names, not a list of 1's and 0's. I went with disease names instead of disease indexes, since I thought that would be clearer to users. `MultiLabelProcessor` converts these names to indexes and outputs a helpful map between the two (see the unit test output snippet below).

```
Generating samples for ChestXray14MultilabelClassification with 1 worker: 100% 10/10 [00:00<00:00, 326.83it/s]
Label labels vocab: {'cardiomegaly': 0, 'effusion': 1, 'emphysema': 2, 'hernia': 3, 'infiltration': 4}
```

pyhealth/datasets/chestxray14.py
EricSchrock: @jhnwu3 if I'm reading the code correctly, `covid19_cxr.py` is for a **multiclass** dataset, however `chestxray14.py` is a **multilabel** dataset. Wouldn't `MIMIC4CXRDataset` from `mimic4.py` be a better comparison?

The current code does construct a `chestxray14-metadata-pyhealth.csv` file (see the `_index_data` function) and then calls the base init process (right after the call to `self._index_data()`). Perhaps `self._data`, `__len__`, and `__getitem__` are causing the confusion? These are only meant for testing (see `test_chestxray14.py`) but looking at `test_bmd_hs.py` I see how I can test without them, so I can delete them. Is this what you're going for?

pyhealth/tasks/chestxray14_multilabel_classification.py
EricSchrock: @jhnwu3 I think the complexity here comes from the fact that this is a **multilabel** dataset, not a **multiclass** dataset. `BMDHSDiseaseClassification` is another multilabel dataset classification task with similar complexity. Am I missing something?

pyhealth/datasets/chestxray14.py
jhnwu3: Yeah, pretty much. 

The image schema and the Task will construct a SampleDataset that will return it in the format image tensor format: https://pyhealth.readthedocs.io/en/latest/_modules/pyhealth/processors/image_processor.html#ImageProcessor


Sorry, was mostly pointing towards Covid19CXR as an example of how to structure the dataset and task code.

pyhealth/tasks/chestxray14_multilabel_classification.py
```diff
@@ -0,0 +1,64 @@
...
+        Returns:
+            List[Dict]: A list containing a single dictionary with:
+                - 'image': path to the chest X-ray image.
+                - 'labels': a list of labels for diseases present in the image (strings from the list ChestXray14Dataset.classes).
+
+        Raises:
+            ValueError: If the number of chestxray14 events is not exactly one.
+        """
+        events: List[Event] = patient.get_events(event_type="chestxray14")
+        if len(events) != 1:
```
EricSchrock: @jhnwu3 currently I'm just treating every X-ray image as a separate "patient". This seems to match most of the other current datasets and tasks. However, ChextX-ray14 does include "Patient ID" and "Follow-up #" fields (see link below). Would you prefer I keep the current "flat" dataset implementation that treats each image as a separate patient or that I update the dataset and tasks to support multiple events per patient?

https://drive.google.com/file/d/1mkOZNfYt-Px52b8CJZJANNbM3ULUVO3f/view?usp=sharing

jhnwu3: Ahh, if it's like MIMIC-CXR in that case. Then, yeah, we should be able to support multiple images per patient. I think it's good to give the user the choice to decide in their own tasks whether they want to do a flattened classification task or a longitudinal form of modeling (i.e some new works in X-ray diffusion try to model the change in a patient's x-ray over time given some condition).

tests/core/test_chestxray14.py
jhnwu3, line 1: ahh I just realized, can you move this into tests/core, so we can run it in our CI/automated unit tests. You may have to pull from master

jhnwu3 (changes requested):

EricSchrock:
Remaining to-dos
1. Determine whether `ChestXray14Dataset` should override `set_task`. Looking at similar datasets, `COVID19CXRDataset` does but `MIMIC4CXRDataset` does not.
2. Update the API docs for the new dataset and tasks.
3. Add an example notebook.

jhnwu3 (changes requested):
One last request, can you add a corresponding .rst file to the docs/api/datasets for this? This way it shows up on the page once we add it to the package.

jhnwu3 (changes requested):
One last change, I should add it into the docs since it's not well-described, but can you add a reference to the datasets.rst file here?

https://github.com/sunlabuiuc/PyHealth/blob/master/docs/api/datasets.rst

After that, happy to merge it in! It's really nice to see it come together like this!

https://github.com/sunlabuiuc/PyHealth/blob/master/docs/api/tasks.rst

same thing for the tasks.rst file here.

PR #432, opened 2025-05-06
Add a note about MIMIC-CXR-JPG in the MIMIC4CXRDataset loader

swghosh:
While working on [Vision-Language Generative Model for View-Specific Chest X-Ray Generation](https://arxiv.org/abs/2302.12172) paper and skimming through [MIMIC-CXR-JPG](https://physionet.org/content/mimic-cxr-jpg/2.0.0/) dataset,
we found that the newly added `MIMIC4CXRDataset` already implements the said dataset and the loader would explicitly expect to read images from the JPG format instead of the DCM format used in the original variant of the dataset (MIMIC-CXR).

To avoid users from running into gotchas due to this, added a note and warning.

---

Contributor Name(s): Swarup Ghosh, Mick Enev
UIUC NetID: menev2, swarupg2.

cc: @jhnwu3, per thread: https://discord.com/channels/1359223185253597335/1359230031074820299/1369305683501256735

also, cc: @MickEnev

### DIFF
--- a/pyhealth/datasets/mimic4.py
+++ b/pyhealth/datasets/mimic4.py
@@ -121,6 +121,11 @@ class MIMIC4CXRDataset(BaseDataset):
     This class is responsible for loading and managing the MIMIC-CXR Chest X-ray dataset,
     which includes tables such as metadata, chexpert, and radiology.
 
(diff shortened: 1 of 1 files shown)


jhnwu3:
Closing this PR as it lacks proper labeling. Please add appropriate labels and reopen if needed.

PR #480, opened 2025-05-08
new use case/example for PyHealth

spot-p:
From: Swaroop Potdar netid (spotd), Jake Cumberland netid (jakepc3)

We are submitting a new use case for PyHealth for MIMICIV. This is a notebook example which users can utilize to follow along.

Summary:
Considering the privacy concerns and paucity of available data, the goal of this workflow is to match the MIMIC-IV Waveform Database (WaveDB) with MIMIC-IV Clinical Ground truth data to enable useful analysis. Specifically, by linking patient diagnosis information with waveform data, we can create datasets that can be used for machine learning tasks. These datasets make for an ideal pilot or proof of concept study, where we can asses the validity of our hypothesis before collecting more and diverse data on a larger scale.
For example, we can identify which patients are positive for conditions like Atrial Fibrillation, and then match them with their corresponding waveform records and choose control subjects who are negative for the target condition but still have waveform records.


spot-p:
I have tried everything to exclude the permission changes from being excluded from the PR, but have been unsuccessful. It looks like the colab environment introduced some unwanted permission changes on the files never touched.

jhnwu3:
Closing this PR as it lacks proper labeling. Please add appropriate labels and reopen if needed.

PR #496, opened 2025-05-08
Adding Basic Saliency Mapping (fitzpa15)

Nimanui:
Adding a function to allow Gradient saliency to be applied to Pyhealth models that work with image data. I have also included a python notebook demonstrating this function being used, and the visual output. This is a bit of a proof of concept that we can apply saliency to some of the Pyhealth models, although would only be the start if there was a desire to add additional saliency techniques (this was the simplest of the ones I tried in my paper replication).


Nimanui:
Need to rebase the main branch, one moment

PR #519, opened 2025-07-12
fix readthedoc dependency error

jhnwu3:
The doc strings aren't updating properly due to the pydantic import that doesn't exist when running the readthedocs pipeline.


zzachw (approved):
LGTM. Thanks!

PR #524, opened 2025-07-19
Switch from Python `setuptools` to Pixi

plandes:
This modernizes the build to [pixi](https://pixi.sh) and currently provides the following:
- Run the unit tests in controlled lock (file) environments
- Build a distribution wheel


jhnwu3 (approved):
Works! Will need to update the metadata to PyHealth 2.0a, but let's keep this rolling!

PR #539, opened 2025-09-09
GitHub Actions Automated Test

muchemicarol:
Included an action to automate running of tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,7 @@ name: CI
 
 on:
   pull_request:
+    branches: [ master ]
   push:
     paths-ignore:
       - '**.md'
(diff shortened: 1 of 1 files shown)


jhnwu3:
Poking @plandes  here, since iirc we have an automated set of tests already setup in the makefile

plandes:
@muchemicarol, @jhnwu3 sorry, but this change should be rejected.  Non-NLP tests should be added as:
    `PyHealth/tests/core/test_<your test name>.py`

If your test has a new requirement that's needed both in the code and the test, append to the `project.dependencies` in the `pyproject.toml` file.

If your test has a new requirement only needed for the test itself, append to the `tool.pixi.feature.test.pypi-dependencies` TOML table in the `pyproject.toml` file.

The tests are run with either `make testall` or `pixi run test && pixi run testnlp`.

muchemicarol:
Hey @jhnwu3, @plandes thanks for the feedback. It was a GitHub Action to run all tests (not a targeted NLP or Non-NLP test).

PR #570, opened 2025-10-01
Fix sample dataset indexing

### LINKED ISSUE #369: Fix sample dataset indexing
We should probably add a check. But, it's mostly so we don't break compatibility with legacy code like split_by_visit() in https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/datasets/splitter.py

_Originally posted by @jhnwu3 in https://github.com/sunlabuiuc/PyHealth/pull/335#discussion_r2059140897_

ram-from-tvl:
**Fixes #369**

- Refactored the `record_to_index` mapping logic in `SampleDataset.__init__()` to explicitly check for `record_id` first, then fall back to `visit_id` for legacy compatibility
- Added proper None checking to prevent errors when samples lack both `record_id` and `visit_id` fields

### DIFF
--- a/pyhealth/datasets/sample_dataset.py
+++ b/pyhealth/datasets/sample_dataset.py
@@ -68,7 +68,10 @@ class SampleDataset(Dataset):
                 self.patient_to_index[patient_id].append(i)
 
             # Create record_to_index mapping (optional)
-            record_id = sample.get("record_id", sample.get("visit_id"))
+            # Check for record_id first, then visit_id for legacy compatibility
+            record_id = sample.get("record_id")
+            if record_id is None:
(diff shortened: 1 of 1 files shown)


ram-from-tvl:
Hi @jhnwu3 
I have added a check. Please review it and let me know if any changes are required.
Thank you!

jhnwu3:
Hey, can you write a test case in tests/core for this? Something that starts with our synthetic MIMIC-III dataset in one of the notebook examples, and then show it runs for a full pipeline where you split by visits instead of patient id? You may have to dive deeper into how we write our tasks.

jhnwu3:
Closed due to inactivity.

PR #577, opened 2025-10-16
Saliency mapping gradient

Nimanui:
Goal of this pull request is to add basic gradient saliency mapping in the appropriate location in pyhealth as well as providing a clean python notebook example with a basic CNN since a CNN is required to display the saliency map. Assuming this looks ok, I have a basic LRP saliency mapping and GradCAM to implement next since I already have basic pytorch implementations of those already.


pyhealth/interpret/methods/saliency.py
jhnwu3, line 1: I'm wondering if we can make this into more of an object-oriented approach here. 

Something like SaliencyMapper(model) 

have a generate() or process(image, label) function to make it more general in some ways.

Nimanui:
Reasonable, there's probably something small I can test without a full CNN, I'll look at the other tests for options.

Nimanui:
I wasn't able to find something without a CNN, but it was simple to create a test with a dummy CNN. Would you like me to make it use a CNN with other pyhealth modules or is this sufficient? Keeping it with something small will keep it running pretty quickly, I noticed some of the unit tests in the broader suite are very slow.

pyhealth/interpret/methods/saliency.py
jhnwu3, line 1: Hmmm, okay, we're really close!

Just one last requests from my end, instead of passing in a dataloader as an initialization. Is it possible to pass in a batch directly? and make it 

compute_saliency_maps(self, batch) instead of what it is currently where it implicitly calls it on the entire dataloader. This way the user knows exactly which batch or images they're doing a saliency map for.


Mostly, because in practice, we don't want to do it across the entire dataset.

Nimanui:
Some github problems, need to clean this up

PR #587, opened 2025-10-26
[Bounty] GNN PyHeatlh 2.0 ver.

Logiquo:
**Contributor:** Yongda Fan (yongdaf2@illinois.edu)

**Contribution Type:** Model, Tests, Example Notebooks

## Description

Added Graph Neural Network (GNN) models including Graph Convolutional Network (GCN) and Graph Attention Network (GAT) to align with PyHealth 2.0 architecture and dataset conventions.

- Implemented full model classes with proper initialization, forward passes, and integration with PyHealth's BaseModel
- Added comprehensive unit tests covering model initialization, forward/backward passes, embedding extraction, and custom hyperparameters
- Created example notebooks for GCN and GAT training on MIMIC-IV dataset, showcasing the Trainer workflow, evaluation metrics, and inference capabilities
- Updated model docstrings to follow Google style guidelines for better documentation

## Files to Review

- `pyhealth/models/gnn.py`
- `tests/core/test_gnn.py`
- `examples/gcn_mimic4.ipynb`
- `examples/gat_mimic4.ipynb`

### DIFF
--- a/examples/gcn_mimic4.ipynb
+++ b/examples/gcn_mimic4.ipynb
@@ -11,10 +11,18 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "id": "6b8378db",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Running on device: cuda\n"
+     ]
+    }
+   ],
    "source": [
     "import os\n",
     "import random\n",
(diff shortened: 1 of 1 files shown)


jhnwu3 (commented):
Looks mostly good to me. Will check back again tomorrow morning for a more conceptual check as I just got back from a long trip. 

Thanks a lot Yongda!

pyhealth/models/gnn.py
jhnwu3, line 1: Ooooh, okay. Some thoughts: 
- I now realized our initial implementation was also imperfect and didn't really consider the various features that are eventually embedded. Tbh, it was pretty barebones.

- We should make the adjacency matrix a required argument in the forward() function as typically we can define graphs in many different ways. Feature vs. feature, patient vs. patient. Given X=[B,F] So if it's batch-wise, our A = [B,B], and if it's feature-wise it's [F,F]. Patient-wise, it's [B,B], and if we're discussing event-wise, it's [N,N] where N Is the max length of the number of events in that batch. 
- For the examples, let's have a test case forward for each type of adjacency matrix.

- It's impressive how fast you're doing this! Thanks again! This is really high quality work! I look forward to our continued collaborations. I think the professor is happy to see these contributions.

jhnwu3: https://medium.com/@jrosseruk/demystifying-gcns-a-step-by-step-guide-to-building-a-graph-convolutional-network-layer-in-pytorch-09bf2e788a51

Here are some additional resources in case you want to see what I mean. Otherwise, I think the rest of it looks good to me.

Logiquo: I've implement the feature-wise and visit-wise. The patient-wise one is trick because

Patient-level graphs are conceptually problematic because each batch only contains a small, incomplete subset of all patients. This means the adjacency matrix built per batch does not represent the full patient network and changes every iteration, depending on how samples are batched.

So i decide to ignore paitent-wise for now.

jhnwu3 (approved):
lgtm!

Sidenote for myself: I think we'll have to extend this with some of Patrick's work later if we actually want to build legitimate patient graphs (but I think this approach is the best way can do so far) as it actually is compatible with our existing dataloaders and training pipeline. Thanks a lot!

PR #593, opened 2025-11-02
[BMS] Deeplift for MIMIC4 Mortality Prediction

Logiquo:

**Contribution Type**: Interpretability, Models, Tests, Example Script, Docs

**Description**
Implemented DeepLIFT interpretability support aligned with PyHealth 2.0 architecture, providing activation hook management and completeness enforcement.
Extended StageNet with DeepLIFT hook registration and embedding utilities so interpretability works across trainer flows and mixed inputs.
Added targeted unit tests verifying DeepLIFT attribution correctness for embedding and continuous pathways, including completeness checks.
Created a StageNet+DeepLIFT example on MIMIC-IV to demonstrate usage, baseline selection, and attribution visualization.
Documented the new interpretability API in the Sphinx reference to expose the DeepLIFT interface.

**Files to Review**
pyhealth/interpret/methods/deeplift.py
pyhealth/models/stagenet.py
tests/core/test_deeplift.py
examples/deeplift_stagenet_mimic4.py
docs/api/interpret/pyhealth.interpret.methods.deeplift.rst

### DIFF
--- a/pyhealth/interpret/methods/deeplift.py
+++ b/pyhealth/interpret/methods/deeplift.py
@@ -611,8 +611,9 @@ class DeepLift(BaseInterpreter):
 
         scale = torch.ones_like(delta_output)
         if total is not None:
-            denom = total.abs()
-            mask = denom > eps
+            # Preserve the sign of the raw contributions by dividing by the signed sum
+            denom = total
+            mask = denom.abs() > eps
             scale[mask] = delta_output[mask] / denom[mask]
 
         for key, contrib in contributions.items():


Logiquo:
image

jhnwu3 (commented):
Is it possible to add a small test_deeplift.py in tests/core?

https://github.com/sunlabuiuc/PyHealth/blob/master/tests/core/test_integrated_gradients.py

This is just to make sure we can always test that the approach will initialize and be callable to some degree of confidence in the environment.

pyhealth/interpret/methods/deeplift.py
jhnwu3, line 1: this is something I need to do a better job of documenting and enforcing down the road, which is can we add an example initialization and usage with some pseudo data in the doc strings here? It's a really useful way for autodocs such that when users look to use your code, it'll be pretty straightforward.

jhnwu3: Overall logic makes sense and is in agreement with the paper description, so nice work! Would be nice to have some quality of life additions.

Logiquo: New docs added.

pyhealth/interpret/methods/deeplift.py
```diff
@@ -0,0 +1,638 @@
...
+    # ------------------------------------------------------------------
+    def _deeplift_embeddings(
+        self,
+        inputs: Dict[str, torch.Tensor],
+        baseline: Optional[Dict[str, torch.Tensor]],
+        target_class_idx: Optional[int],
+        time_info: Dict[str, torch.Tensor],
+        label_data: Dict[str, torch.Tensor],
+    ) -> Dict[str, torch.Tensor]:
+        """DeepLIFT for discrete inputs operating in embedding space.
```
jhnwu3, line 402: Just curious, is time_info required? It's probably jank programming practice, but can we make this an optional parameter for compatibility with say something like an MLP or an RNN later? https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/models/mlp.py

pyhealth/interpret/methods/deeplift.py
```diff
@@ -0,0 +1,638 @@
...
+        mapped: Dict[str, torch.Tensor] = {}
+        for key, contrib in contribs.items():
+            mapped[key] = contrib.to(device)
+        return mapped
+
+    # ------------------------------------------------------------------
+    # Utility helpers
+    # ------------------------------------------------------------------
+    @staticmethod
+    def _determine_target_index(
```
jhnwu3, line 561: hmm, we might have to implement that function as a Base function later in an interpretability module refactor. 

Since this type of function is very commonly called across all interpretability approaches.

jhnwu3 (changes requested):
Yeah, I think my only current concern is maybe making it compatible with an MLP class just in case we want to extend DeepLift beyond StageNet (i.e not assuming a time tensor input always). 

But, lgtm beyond that. I think once we compile everyone's approaches, we can probably standardize a parent class for this.

pyhealth/interpret/methods/deeplift.py
Logiquo: I've changed it optional.

jhnwu3 (approved):
lgtm! 

Should work for all foreseeable approaches.

PR #594, opened 2025-11-02
Basic gradient saliency visualization

Nimanui:
Had to refresh the branch due to some github shenanigans. Should be the same as the previous pull request, but I completed adjustments to let it render and manage batches instead of inputting a dataloader (I left the dataloader stuff if someone wants to do a larger set but perhaps it's redundant). I've also updated the example python notebook accordingly and refactored the tests. Also did some light refactoring to emphasize that this is just a basic pytorch gradient saliency visualization class.

### DIFF
--- a/docs/api/interpret/pyhealth.interpret.methods.basic_gradient.rst
+++ b/docs/api/interpret/pyhealth.interpret.methods.basic_gradient.rst
@@ -4,95 +4,12 @@ pyhealth.interpret.methods.basic_gradient
 Overview
 --------
 
-The ``BasicGradientSaliencyMaps`` class provides gradient-based saliency map visualization for 
-PyHealth's image classification models. This interpretability method helps identify which regions 
-of medical images most influenced the model's prediction by computing gradients of model outputs 
-with respect to input pixels.
+The BasicGradientSaliencyMaps method computes gradient-based saliency maps for PyHealth's image
+classification models. This helps identify which regions of medical images most influenced the
+model's prediction by visualizing gradients of model outputs with respect to input pixels.
 
-This method is particularly useful for:
-
-- **Clinical interpretability**: Understanding which image regions drove a particular diagnosis
-- **Model debugging**: Identifying if the model focuses on clinically relevant features
-- **Trust and transparency**: Providing visual explanations for model predictions
-- **Error analysis**: Comparing saliency maps for correct vs. incorrect predictions
-
-The implementation computes saliency by taking the maximum absolute gradient across color channels
-for each pixel, highlighting the most influential regions in the input image.
-
-Key Features
-------------
-
-- **Dual input support**: Process batches from DataLoader or direct batch inputs
-- **Flexible visualization**: Matplotlib overlay with configurable transparency
-- **Label comparison**: Display both true labels and model predictions
-- **Efficient storage**: Separate caching for different data sources
-
-Usage Notes
------------
-
-1. **Gradients required**: Do not use within ``torch.no_grad()`` context
-2. **Model compatibility**: Works with PyHealth image classification models
-3. **Memory usage**: Limit batch count to control memory consumption
-4. **Batch visualization**: Use ``batch_index`` for pre-computed maps, omit for on-the-fly computation
-
-Quick Start
------------
-
-.. code-block:: python
-
-    from pyhealth.interpret.methods.basic_gradient import BasicGradientSaliencyMaps
-    from pyhealth.datasets import get_dataloader
-    import matplotlib.pyplot as plt
-
-    # Assume you have a trained image model and dataset
-    model = TorchvisionModel(dataset=sample_dataset, ...)
-    # ... train the model ...
-
-    # Create interpretability object with dataloader
-    dataloader = get_dataloader(dataset, batch_size=32, shuffle=True)
-    saliency_maps = BasicGradientSaliencyMaps(
-        model=model,
-        dataloader=dataloader,
-        batches=3
-    )
-    saliency_maps.init_gradient_saliency_maps()
-
-    # Visualize from pre-computed maps
-    saliency_maps.visualize_saliency_map(
-        plt,
-        image_index=10,
-        batch_index=0,
-        title="Gradient Saliency",
-        id2label={0: "Normal", 1: "COVID", 2: "Pneumonia"},
-        alpha=0.6
-    )
-
-Custom Batch Example
---------------------
-
-.. code-block:: python
-
-    import torch
-
-    # Create a custom batch (e.g., filter by diagnosis)
-    covid_samples = [s for s in dataset.samples if s['disease'].item() == covid_label]
-    covid_batch = {
-        'image': torch.stack([covid_samples[i]['image'] for i in range(32)]),
-        'disease': torch.stack([covid_samples[i]['disease'] for i in range(32)])
-    }
-
-    # Initialize with custom batch
-    saliency_maps = BasicGradientSaliencyMaps(model=model, input_batch=covid_batch)
-    saliency_maps.init_gradient_saliency_maps()
-
-    # Visualize (no batch_index means use input_batch)
-    saliency_maps.visualize_saliency_map(
-        plt,
-        image_index=0,
-        title="COVID Sample",
-        id2label=id2label,
-        alpha=0.6
-    )
+For a complete working example, see:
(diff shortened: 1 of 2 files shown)


jhnwu3:
Running unit tests, if it passes, I think it looks pretty good to me. We can always iterate further later. I think at some point, we might need a standardized API for the interpretability module haha since I see a lot of similarities in our approaches here.

jhnwu3 (approved):
lgtm

pyhealth/interpret/methods/basic_gradient.py
jhnwu3, line 1: Ahh, sorry, I should've reviewed more in detail earlier. Was stuck in a meeting.

Some quick thoughts that I think would make more sense in terms of a refactor here again (sorry to make you do this, but I think it would reduce the complexity of the approach.

1. We shouldn't store the dataloader inside of the class as it's really heavy to have an entire dataloader inside of an interpretability class here. The idea here is we simply call this class's "attribute" or .saliency_map(batch) as we're iterating across the dataloader. This way, technically the user can define whether or not to do it across the entire dataset themselves or decide how they want to do the sampling of interpretation.

2. If possible, the .attribute() or .process_batch() would just return the tensor attributions themselves as I think that would be a pretty straightforward way of letting the user use these attribution features.

3. The reason why we want to return the raw attribution tensor features is because a lot of downstream interpretability methods (like adversarial learning approaches) can often use these attributions in other creative ways.

jhnwu3 (changes requested):

pyhealth/interpret/methods/basic_gradient.py
jhnwu3: Can we also have it inherit from the new master's BaseInterpreterClass?

Nimanui: I could try but this already far more overengineered than the intent of just a basic method that just does a very simple gradient on input image data from a CNN. Would it be possible to come back to that on a later pass and then if this is helpful I can expand it to support StageNet at that point?

Nimanui: I also would like to try to use some minor image support with the LRP but if I reference any of this code or the example in it, it becomes dependent on this and a bit of a problem. I'm happy to come back to this later but I'm kind of stuck with this still being open.

jhnwu3: Yes exactly, I think we can definitely simplify this class substantially. There are some things I'm a little confused about. Let me know your thoughts.

1. We can remove the input_batch init argument here? Do we need an assertion check here? 
2. It feels like _process_batch and attribute() are redundant here?
3. I also feel like _compute_saliency() isn't called anywhere? Or get_saliency_maps()?
4. I really like the visualize function idea

jhnwu3 (changes requested):
Oh on the topic of documentation, can you add this file reference to: https://github.com/sunlabuiuc/PyHealth/blob/master/docs/api/interpret.rst

pyhealth/interpret/methods/basic_gradient.py
Nimanui: Good thoughts, and reasonable questions. 

1. The goal of the input batch is to allow the user to provide a specific set of pictures/raw data that they want to process through the interpretability process. Otherwise I would have to add a large quantity of sorting or other tools to figure out which data they want to use, or go back to the original implementation which just processed all of it (with some options to restrain it a bit for performance). Maybe I'm misunderstanding the base class, but it doesn't feel like we currently have a consistent way to help users decide what data they want to look at (perhaps doubly important with the visualization applictations).
2. You are correct that there is some functional overlaps there, I'll merge them a bit to reduce the redundancy
3. Good call, I think it may have gotten lost in the shuffle, I think the get is being called in the python notebook example but may also be extra.
4. I think I'll try to refactor that into a general class in a future iteration, part of the push to get this through is that you can visualize LRP and it's quite a fun example, but it mildly depends on those functions here. In particular having the base image underneath the saliency map makes for an excellent visual (imo).

Nimanui: Ok, found a few more spots, part of the issue is when I devoloped this I figured it would be better to build a dictionary that has the input image, saliency map, and label all kept together. Otherwise, if we only store the saliency map on it's own, it will quickly become impossible to tie it back to the original image for the visualization (and well they are somewhat helpful on their own, I really like the idea of having the original image underneath). So the _compute_saliency function is just setting up that dictionary and constructing it. I agree that the getter is a bit redundant, that was originally because I had it using a lazy initialization where it didn't compute all of the saliency maps until they were specifically asked for. This made more sense when it had the "process the entire inputted dataset" option originally and isn't as helpful now that it is only doing processing by batch anyway. I'll tie the initialize into the main init and just leave a simple getter for testing use if someone wants to grab the saliency dictionary.

jhnwu3: 1. So the attribute(**data_batch) should be able to let users pass in any batch and return their attributions here. For reference, you can iterate through a pytorch dataloader and filter for a specific input. I think my concern is that it doesn't make sense for it to be in the init function? When other functions clearly can perform the attribution on any batch the user decides?



4. (4) No worries, we can leave those be. Happy to have them as they are now.

Nimanui: 1. At that point then, isn't it just a static function and class at that point then? I'm not sure what you would store from in the class if you don't even have the processing batch on init. I could make it optional for the existing application (which only calls attribute as a helper method), and I'll adjust it so it updates the batch if one is passed with the attribute method. I think that lines it up a bit closer.
2. Turns out it slimmed up real nice, thank you for poking about this. I was able to clear about 100 lines and I'm pretty happy with the results.

jhnwu3: 1. The main thing I think (and feel free to disagree with me on this) is just making things standardized where we can assume that every (local) interpretability method should have an attribute() function that can be leveraged in our interpretability metrics here since the metrics assume you have some type of function attribute(), and it's nice to centralize the logic around this abstract interpreter class.

There's also things like not being able to compute gradients on discrete variables (see the DeepLift and IntegratedGradient examples) where you'll need some type of specialized logic here. I would imagine if these gradient-based saliency maps were computed on things like discrete word tokens here. It would require a little more complicated logic here, whether that's storing different types of baseline inputs to compute some difference in activations or just being able to have different modes (i.e SHAP can be computed a variety of different ways, but at the end of the day, they're all estimating the true SHAPley values)

But, honestly, I think object-oriented approaches are just a little cleaner in terms of abstraction and code readability (for me personally) Technically if we were to follow a pythonic style, we could pass around a bunch of functions everywhere, but imo that would be really hard to read.

Nimanui: Ok, those changes are in as well, let me know if you find anything else. There is probably still a few minor refactoring pieces I can think of but I think that's all of the major ones. If there are additional examples we want to add I'd be inclined to do that later.

Nimanui: And the notebook is up to date with this code, although again I'm not sure if we want to do some more notebooks with different core models. I only picked this one as it was an easy fit, but I think we could add it to the chestxray 14 multiclassifier as well. There are also plenty of fun options for adding visualization tools, again thoughts for later work.

docs/api/interpret/pyhealth.interpret.methods.basic_gradient.rst
jhnwu3, line 1: For the sake of vibe-coding purposes, can we move these details into the docstrings in the class implementation? Also, it seems the docs are out of date. The nice part about having it specifically in the doc strings (that get rendered by autodocs) is that the LLM will update the doc strings and the logic every iteration so it'll be easy to keep things maintained.

I think the examples/notebook might be out of date too?

Otherwise, I think we're pretty much there. Thank you for taking the time to iterate on this.

Nimanui: The notebook is updated (I've been testing with it for every change in addtion to the unit tests), I'll poke the documentation and let you know, that probably is out of date. I don't mind moving it, the other documentation I just added is up to date.

Nimanui: The rst is updated to resemble the other interpretability ones (sorry I didn't do that sooner) and I added the gist of what it had to the doc string in the code. Excellent call there and thank you for helping me make this much better.

PR #596, opened 2025-11-05
[Bounty] MoleRec

lehendo:
Arjun Chatterjee
netID: arjunc4

Model implementation

Attempt at implementing the MoleRec model into PyHealth
Updated to current BaseModel/processor API
Added unit tests (small synthetic data)
Made a Jupyter notebook that trains and evaluates MoleRec on MIMIC-III

Additions to pyhealth/models/molerec.py:
BaseModel compatibility
Processor -> tokenizer compatibility
Forward pass improvements (using kwargs inputs)
Molecular/SMILES increased robustness
Substructure mask and average projection generation
GIN-based molecular graph encoding
Attention-based aggregation for drug recommendations
DDI-aware loss

Files to review:
pyhealth/models/molerec.py - Full MoleRec implementation
tests/core/test_molerec.py - Main testing suite (unit tests)
examples/molerec_mimic3.ipynb - Jupyter notebook for training and evaluating on MIMIC-III
pyproject.toml - Added ogb dependency (make sure you install ogb first)

### DIFF
--- a/pyhealth/models/molerec.py
+++ b/pyhealth/models/molerec.py
@@ -765,11 +765,8 @@ class MoleRec(BaseModel):
         procedures = embedded["procedures"]
         
         # [batch, visits, embedding_dim]
-        conditions = conditions.sum(dim=2)
-        procedures = procedures.sum(dim=2)
-
-        conditions = self.dropout_fn(conditions)
-        procedures = self.dropout_fn(procedures)
+        conditions = torch.sum(self.dropout_fn(conditions), dim=2)
+        procedures = torch.sum(self.dropout_fn(procedures), dim=2)
         
         # [batch, visits, hidden_dim]
         condition_emb, _ = self.rnns["conditions"](conditions)


pyhealth/models/molerec.py
jhnwu3, line 1: Sorry for the late review. 

We'll need to get rid of the SampleEHRDataset here as that is now defunct. Basically, we should assume that the model here takes SampleDataset instead as the type of dataset.  Everything else looks roughly good to me. Just need to make sure the typing is correct. Also if possible, can you add docstrings to show how to use the model? 

Specifically, like how we do it in StageNet here?

https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/models/stagenet.py

It's mostly so the autodocs can update with a cleaner example such that a user navigating the site can easily figure out how to use the model.

jhnwu3 (changes requested):

pyhealth/models/molerec.py
```diff
@@ -632,69 +737,98 @@ def generate_average_projection(self) -> Tuple[torch.Tensor, List[str]]:
...
+        embedded = self.embedding_model(kwargs)
+        
+        # embedded["conditions"] shape: [batch, visits, codes_per_visit, embedding_dim]
+        # embedded["procedures"] shape: [batch, visits, codes_per_visit, embedding_dim]
+        conditions = embedded["conditions"]
+        procedures = embedded["procedures"]
+        
+        # [batch, visits, embedding_dim]
+        conditions = conditions.sum(dim=2)
+        procedures = procedures.sum(dim=2)
```
jhnwu3: I think the dropout order might be incorrect here (?). 
https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/models/molerec.py#L641
`        embeddings = torch.sum(self.dropout_fn(embeddings), dim=2)`

If we look at the old implementation in their encoding. It seems they call dropout before summing, which changes how much randomness occurs during a forward pass.

jhnwu3 (changes requested):

PR #597, opened 2025-11-05
Add early stopping patience parameter to Trainer

Razin93937:
**Contributor:** Sayeed Sajjad Razin ([razin93937@gmail.com](mailto:razin93937@gmail.com))
**Contribution Type:** Training, Tests

**Description**
Added an optional patience parameter to the Trainer class for early stopping. If the monitored validation metric doesn’t improve for patience epochs, training stops early. When patience=None (default), training runs for all epochs (backward-compatible).

Added tests to verify:
- Early stop triggers when validation worsens
- No stop when metrics keep improving
- Disabled behavior when patience=None

**Files to Review**
pyhealth/trainer.py
tests/core/test_early_stopping.py

### DIFF
--- a/pyhealth/trainer.py
+++ b/pyhealth/trainer.py
@@ -160,6 +160,7 @@ class Trainer:
         logger.info(f"Monitor: {monitor}")
         logger.info(f"Monitor criterion: {monitor_criterion}")
         logger.info(f"Epochs: {epochs}")
+        logger.info(f"Patience: {patience}")
 
         # set optimizer
         param = list(self.model.named_parameters())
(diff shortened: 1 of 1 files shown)


tests/core/test_early_stopping.py
jhnwu3, line 1: Any idea how long this takes to run? I see 50 epochs, which could be problematic here.

Razin93937: Thanks for pointing that out! I originally overlooked the test runtime. I’ve updated the tests to run in about 0.5s by reducing the epochs to 10 and setting the embedding and hidden dimensions to 1. The change is now committed.

pyhealth/trainer.py
jhnwu3, line 1: Okay, i've taken a deeper look. Just one last remark. Is it possible to add a quick logger info with the rest of the other hyperparameters for logging purposes? Just so the user when they run this with patience, it will be recorded somewhere. Other than that, nice work! Will merge it end of the day hopefully after this change.

jhnwu3 (changes requested):

pyhealth/trainer.py
Razin93937: Thanks for the review! I've added logging for the patience parameter.

**Files to Review**
`pyhealth/trainer.py`

PR #600, opened 2025-11-07
Dataset: (BMD-HS) Heart Sound Recordings for Automated Cardiovascular Disease Diagnosis

Razin93937:
**Contributor:** Sayeed Sajjad Razin ([razin93937@gmail.com](mailto:razin93937@gmail.com))
**Contribution Type:** Dataset 

**Dataset**
[BMD-HS (BUET Multi-disease Heart Sound](https://arxiv.org/abs/2409.00724)

**Description**
The BMD-HS dataset contains over 800 phonocardiogram (PCG) recordings collected from patients at the National Institute of Cardiovascular Disease (Dhaka). Recordings span six clinically relevant categories: Aortic Stenosis (AS), Aortic Regurgitation (AR), Mitral Regurgitation (MR), Mitral Stenosis (MS), and Normal (N).
The dataset provides multi-label annotations, and includes demographic metadata (age, gender, smoking status, and living region). Each patient contributes up to eight 20-second recordings sampled at 4 kHz. The eight recordings come from different auscultation points (aortic valve, pulmonic valve, tricuspid valve, and mitral valve) and patient positions (supine and sitting).

**Files to Review**
`docs/api/datasets/pyhealth.datasets.BMDHSDataset.rst`
`pyhealth/datasets/__init__.py`
`pyhealth/datasets/bmd_hs.py`
`pyhealth/datasets/configs/bmd_hs.yaml`
`tests/core/test_bmd_hs.py`

### DIFF
--- a/pyhealth/tasks/bmd_hs_disease_classification.py
+++ b/pyhealth/tasks/bmd_hs_disease_classification.py
@@ -1,7 +1,6 @@
-from typing import Any, Dict, List, Optional
-
-from pyhealth.processors.audio_processor import AudioProcessor
+from typing import Any, Dict, List
 from .base_task import BaseTask
+
 
 class BMDHSDiseaseClassification(BaseTask):
     """Multi-label classification task for heart valve diseases.
@@ -26,51 +25,34 @@ class BMDHSDiseaseClassification(BaseTask):
         task_name (str): The name of the task, set to "BMDHSDiseaseClassification".
         input_schema (Dict[str, str]): The input schema specifying the required
             input format. Contains:
-            - "recording_1" through "recording_8": "tensor"
+            - "recording_1" through "recording_8": "audio"
             - "age": "float"
             - "gender": "categorical"
             - "smoker": "binary"
             - "lives": "binary"
-        output_schema (Dict[str, tensor]): The output schema specifying the output
+        output_schema (Dict[str, str]): The output schema specifying the output
             format. Contains:
             - "diagnosis": "multilabel"
     """
 
     task_name: str = "BMDHSDiseaseClassification"
-    
+
     input_schema: Dict[str, str] = {
         # 8 heart sound recordings
-        "recording_1": "tensor",
-        "recording_2": "tensor",
-        "recording_3": "tensor",
-        "recording_4": "tensor",
-        "recording_5": "tensor",
-        "recording_6": "tensor",
-        "recording_7": "tensor",
-        "recording_8": "tensor",
+        **{f"recording_{i}": "audio" for i in range(1, 9)},
         # Patient metadata
         "age": "regression",
         "gender": "binary",
         "smoker": "binary",
         "lives": "binary",
     }
-    
+
     output_schema: Dict[str, str] = {
         # Multi-label disease classification
         "diagnosis": "multilabel",
     }
-    
-    def __init__(self,
-                 sample_rate: Optional[int] = 4000, # BMD-HS original sample rate
-                 duration: Optional[float] = 20.0, # maximum duration of recordings in BMD-HS
-                 to_mono: bool = True,
-                 normalize: bool = True,
-                 mean: Optional[float] = None,
-                 std: Optional[float] = None,
-                 n_mels: Optional[int] = None,
-                 n_fft: int = 400,
-                 hop_length: Optional[int] = None,              
-                 ) -> None:
+
+    def __init__(self) -> None:
         try:
             import soundfile
         except ImportError:
@@ -79,17 +61,6 @@ class BMDHSDiseaseClassification(BaseTask):
                 "Install it with: pip install soundfile"
             )
         super().__init__()
-        self.audio_processor = AudioProcessor(
-            sample_rate=sample_rate,
-            duration=duration,
-            to_mono=to_mono,
-            normalize=normalize,
-            mean=mean,
-            std=std,
-            n_mels=n_mels,
-            n_fft=n_fft,
-            hop_length=hop_length,
-        )
 
     def __call__(self, patient: Any) -> List[Dict[str, Any]]:
         """Process a patient's data to extract features and labels.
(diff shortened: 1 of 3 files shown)


Razin93937:
The new commits add support for the BMD-HS dataset, including a new `AudioProcessor` for handling audio data and a `BMDHSDiseaseClassification` task for multi-label heart valve disease classification.

**Files to Review**
`pyhealth\processors\__init__.py`
`pyhealth\processors\audio_processor.py`
`pyhealth\tasks\__init__.py`
`pyhealth\tasks\bmd_hs_disease_classification.py`
`pyhealth\datasets\bmd_hs.py`
`tests\core\test_audio_processor.py`
`docs\api\processors\pyhealth.processors.AudioProcessor.rst`
`docs\api\tasks\pyhealth.tasks.bmd_hs_disease_classification.rst`

pyhealth/tasks/bmd_hs_disease_classification.py
```diff
@@ -0,0 +1,136 @@
...
+                - "diagnosis": A dictionary with disease labels (AS, AR, MR, MS)
+
+        Raises:
+            AssertionError: If the patient has more than one event.
+        """
+        
+        # Extract recording filenames
+        rec_event = patient.get_events(event_type="recordings")[0]        
+        recordings = {
+            f'recording_{i}': self.audio_processor.process(
```
jhnwu3: Hey, this is a really good PR! Just one major request here: for processing, you'll notice the processing eventually gets done in the build() SampleDataset() class: https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/datasets/sample_dataset.py



So you can actually directly input the processor in the input_schemas like: https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/tasks/medical_coding.py
here where we assume a TextProcessor here for processing it into string inputs.

pyhealth/tasks/bmd_hs_disease_classification.py
```diff
@@ -0,0 +1,136 @@
...
+        output_schema (Dict[str, tensor]): The output schema specifying the output
+            format. Contains:
+            - "diagnosis": "multilabel"
+    """
+
+    task_name: str = "BMDHSDiseaseClassification"
+    
+    input_schema: Dict[str, str] = {
+        # 8 heart sound recordings
+        "recording_1": "tensor",
```
jhnwu3: Recording can be assumed to be AudioProcessor() for the schema. https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/tasks/sleep_staging_v2.py

For instance, this assumes a signal processor. Note: you can directly call the class too like in this medical coding example: https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/tasks/medical_coding.py or its alias in the schema.

Razin93937: Hi, thanks for the helpful review!

One thing I wanted to clarify: I didn’t see a way to pass processor parameters through `input_schema`, since `SampleDataset._get_processor_instance()` only supports no-arg instantiation (either via alias or the processor class). BMD-HS needs dataset-specific and use case-specific audio settings (sample rate, duration, spectrogram on/off, etc.), so I couldn’t use the default `AudioProcessor()` from the schema.

This is why I initialize the processor inside the task’s `__init__`, where the user can directly set parameters. If there is a built-in way to pass processor kwargs through the schema, I’m happy to switch.

Also just joined the Discord (razin93937) and would love to have a discussion!

jhnwu3: Hey yeah sorry, will add more details and links when I get back.

If you look in SampleDataset, it will call all of the processors call functions to do the correct transformations for each of the schemas passed. So no worries about doing an explicit call yourself. 

It's not immediately obvious that this occurs so that's understandable. The tensor schema here means you basically transform it into a tensor twice (first with your specialized transforms and second just calling tensor(tensor)

Razin93937: Thanks so much for the clarification and for the pointers!

One small thing I wanted to flag: BMD-HS really does need some dataset-specific audio processor settings, namely, the sample rate is 4 kHz and clip duration is 20 s. So I initially set those in the task’s `__init__` to create the audio processor for BMD-HS specifically:

```python
# pyhealth/tasks/bmd_hs_disease_classification.py
def __init__(self,
                 sample_rate: Optional[int] = 4000, # BMD-HS original sample rate
                 duration: Optional[float] = 20.0, # maximum duration of recordings in BMD-HS
                 ...          
                 ) -> None:
        ...
        super().__init__()
        self.audio_processor = AudioProcessor(
            sample_rate=sample_rate,
            duration=duration,
            ...
        )
```

If there’s a built-in way to assign processor arguments through the `input_schema`, I’d love a pointer to an example and I’m happy to switch.

If not, would you prefer that I change the default `AudioProcessor` parameters (default is `sample_rate = 16000` and `duration = None`) to match BMD-HS (`sample_rate = 4000` and `duration = 20`) so it works for this dataset? Then I can switch to

```python
# pyhealth/tasks/bmd_hs_disease_classification.py
recordings = {
            f'recording_{i}': 'audio',
            ...
```
I’m totally fine with either approach, just want to align with your conventions.

jhnwu3: Oh, yeah for now, let's just set the default to 4k and duration=20. I now realize we can't do something like

input_schema { "recording1": AudioProcessor(sampling_rate=4000)}  in our current setup, which would let people variably change their schemas. That'll need a patch. Give us sometime to remedy that, but for now we can probably assume a sampling rate and duration for the processor itself. That's a fair point.

Razin93937:
Thanks for confirming, really appreciate you taking the time to dig into it! I’ve updated the default `AudioProcessor` to use `sample_rate=4000` and `duration=20` so the schema path works cleanly for BMD-HS (and updated the tests accordingly). Having schema-level processor arguments on the roadmap sounds perfect, and if it would be helpful, I could potentially take a stab at that in a separate PR.

**Files to Review**
`pyhealth\processors\audio_processor.py`
`tests\core\test_audio_processor.py`
`pyhealth\tasks\bmd_hs_disease_classification.py`

jhnwu3:
Hey, yeah if you want to give it a shot. Would be happy to take the help and give a review! I still need to merge your trainer changes.

PR #604, opened 2025-11-11
[Bounty] ContraWR PyHealth 2.0

Razin93937:
Contributor: Sayeed Sajjad Razin ([razin93937@gmail.com](mailto:razin93937@gmail.com))
Contribution Type: Update to Model, task, Dataset, Test

**Description**

Modifying ContraWR to align with PyHealth 2.0 architecture and dataset conventions. Although this is only indirectly related, I began by updating the [Sleep-EDF Database](https://www.physionet.org/content/sleep-edfx/1.0.0) integration because existing ContraWR test & example seem to rely on it and were not compatible with the new structure. I’d like an intermediate review to confirm whether this is the right direction before moving forward, as I’m not entirely sure whether updating Sleep-EDF should be part of this PR.

**Files to Review**
`pyhealth\datasets\configs\sleepedf_cassette.yaml`
`pyhealth\datasets\sleepedf.py`
`pyhealth\tasks\sleep_staging_v2.py`


Razin93937:
Completed modifying ContraWR to align with PyHealth 2.0 architecture and dataset conventions - including tests and examples.

Files to Review
`pyhealth\datasets\configs\sleepedf_cassette.yaml`
`pyhealth\datasets\configs\sleepedf_telemetry.yaml`
`pyhealth\datasets\sleepedf.py`
`pyhealth\tasks\sleep_staging_v2.py`
`pyhealth\models\contrawr.py`
`tests\core\test_contrawr.py`
`examples\contrawr_sleepedf.ipynb`

pyhealth/datasets/configs/sleepedf_cassette.yaml
jhnwu3, line 1: Just a quick thought:

Not sure if we need to split these into 2 separate yaml files. We should be able to specify both with a file_path metadata change to point to both metadata tables. 

Like instead of "recordings", we can have "cassette" and "telemetry" in the same .yaml file, so users can specify which tables they want to load in their dataset here.
https://www.physionet.org/content/sleep-edfx/1.0.0/
Kind of like how MIMIC has multiple tables in the same .yaml file.

Razin93937: Yeah that does sound better! I made the necessary changes in 9d4ffb67e24363aa72531275ef8a6709bb678bb2

PR #616, opened 2025-11-16
Example notebook with list of Datasets

dalloliogm:
Hi, 
this is an example notebook with a list of datasets mentioned in the documentation. 

I'm going to be honest and admit that I asked an LLM to search over all the files, and draft this notebook. However, I reviewed all the entries, and they seem correct to me.


dalloliogm:
Thanks for the feedback! I've made some changes - do they make any sense? Is this correct?

PR #601, opened 2025-11-07
init commit for testing memory streaming, honestly this is all vibe c…

jhnwu3:
…oded, will need to test and refine extensively for actual release
This pull request introduces a comprehensive streaming mode for large-scale EHR datasets, significantly reducing memory usage by loading patient data from disk on demand. The changes add streaming support to the dataset infrastructure, implement disk-backed patient-level caching with efficient indexing, and provide new APIs for memory-efficient iteration. Additionally, a benchmarking script is added to measure streaming performance and memory efficiency.

**Streaming Mode Infrastructure:**

* Added `stream` and `cache_dir` parameters to dataset initialization, enabling disk-backed streaming mode for large datasets. Streaming mode uses a cache directory to store and retrieve patient data efficiently, keeping peak memory usage under 2GB regardless of dataset size. [[1]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR116-R117) [[2]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR127-R131) [[3]](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR141-R151)
* Implemented `_setup_streaming_cache` and `_build_patient_cache` methods in `BaseDataset` to create a patient-level Parquet cache and an index for fast lookups, using Polars' streaming execution for minimal memory footprint.

**Streaming Patient Iteration:**

* Added `iter_patients_streaming()` method to `BaseDataset`, allowing memory-efficient iteration over patients by loading one patient at a time from disk. Supports optional patient filtering and background preloading for reduced latency.
* Updated `iter_patients()` to raise an error if called in streaming mode, ensuring users use the correct API for memory efficiency.

**API and Import Updates:**

* Exposed `IterableSampleDataset` in the `pyhealth.datasets` module for compatibility with streaming workflows.

**Benchmarking and Documentation:**

* Added `examples/benchmark_streaming.py`, a script to benchmark streaming mode performance on the MIMIC-IV StageNet mortality prediction task, reporting memory usage, processing time, cache size, and speedup over normal mode.
[...]

PR #620, opened 2025-11-19
[Memory] Fix large memory usage during __init__ call. 

Logiquo:
**Contributor**: Yongda Fan ([yongdaf2@illinois.edu](mailto:yongdaf2@illinois.edu)), John Wu

**Contribution Type**: Datasets

**Description**
Fix the large memory usage during dataset's `__init__` call. 

The PR #601 has been used as a reference.

### DIFF
--- a/pyhealth/datasets/base_dataset.py
+++ b/pyhealth/datasets/base_dataset.py
@@ -281,7 +281,7 @@ class BaseDataset(ABC):
 
         # Flatten attribute columns with event_type prefix
         attribute_columns = [
-            pl.col(attr).alias(f"{table_name}/{attr}") for attr in attribute_cols
+            pl.col(attr.lower()).alias(f"{table_name}/{attr}") for attr in attribute_cols
         ]
 
         event_frame = df.select(base_columns + attribute_columns)


Logiquo:
Before this fix:
image

Logiquo:
After this fix:
image

Logiquo:
Note: this does not fix the memory issue when we set task (that would still blow up the memory, and will be fix the subsequence PRs)

Logiquo:
@jhnwu3, this lowername op seems to have no effect?
```Python
attribute_columns = [
    pl.col(attr.lower()).alias(f"{table_name}/{attr}") for attr in attribute_cols
]
```
changes it back to upper name anyways?

examples/memtest.py
```diff
@@ -0,0 +1,32 @@
...
+        if m > PEAK_MEM_USAGE:
+            PEAK_MEM_USAGE = m
+        time.sleep(0.1)
+
+threading.Thread(target=track_mem, daemon=True).start()
+print(f"[MEM] start={PEAK_MEM_USAGE / (1024**3)} GB")
+
+# %%
+from pyhealth.datasets import MIMIC4Dataset
+DATASET_DIR = "/home/logic/physionet.org/files/mimiciv/3.1"
```
dalloliogm, line 19: Hi, can you encode the dataset_dir using the `getpass.getuser()`, instead of hard-coding it to your path? Thanks :-)

PR #624, opened 2025-11-21
Implemented a Base image dataset class

dalloliogm:
Hi,
I was working on the VAE bounty. That bounty is based on the COVID-19 dataset.

While looking at it, I wondered how users could customize the dataset, to use other images than this specific dataset.

So, I took the liberty of creating a BaseImageDataset class, and use that as based for the COVID19 dataset class.

I've also converted `examples/covid19cxr_conformal.py` into a notebook, as I thought that would be easier to read and fit better in an examples/ folder.

There are also a couple of test files for the new dataset class.

I've also improved the error message when the data is not found, suggesting users where they can download it.

I've added openpyxl as dependency because it is needed to read the metadata files.

### DIFF
--- a/pyhealth/datasets/base_image_dataset.py
+++ b/pyhealth/datasets/base_image_dataset.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Dict, Optional
 
 from pyhealth.datasets.base_dataset import BaseDataset


pyhealth/datasets/base_image_dataset.py
jhnwu3, line 1: Hey I think you might be misunderstanding how these BaseDatasets work here.

Namely, if you're doing a VAE and you assume you're going from image to an image, that's a Task.

Datasets are technically distinct from ML tasks as they can often contain many different ML Tasks.

Please see here for https://colab.research.google.com/drive/1kKkkBVS_GclHoYTbnOtjyYnSee79hsyT?usp=sharing

For more tutorials https://pyhealth.readthedocs.io/en/latest/tutorials.html

Let me know if that makes sense.

jhnwu3 (changes requested):

joshuasteier:
Hi @dalloliogm, thank you! As John noted, the VAE use case is better modeled as a Task rather than a new BaseDataset class, since datasets in PyHealth are distinct from ML tasks. This PR has been inactive since November and has merge conflicts, so I'm closing it to keep things tidy. If you'd like to revisit with a Task-based approach, the tutorials John linked are a great starting point — happy to review a new PR.

PR #627, opened 2025-11-27
Bounty: VAE

dalloliogm:
This is the PR for the "VAE" bounty.

Summary of changes:
- some quality of life improvements for COVID19CXRDataset, checking that data exists and expanding the path (e.g. ~/Downloads -> /home/user/Download)
- Added openpyxl requirements as it is needed to read the COVID-19 metadata file, which is Excel
- Implemented options for VAE, including image and timeseries
- Created a new notebook `examples/chestXray_image_generation_VAE.ipynb` to show example of using image-based VAE. Replaced a previous script that did not work
- Improved notebook on time series VAE as well `examples/timeseries_mimic4.ipynb`
- Converted examples/covid19cxr_conformal.ipynb to a notebook and fixed data loading code in examples/ChestXray-image-generation-GAN.ipynb (these are outside of the bounty, but they were broken and the code was similar enough to the VAE examples, so I fixed it)

### DIFF
--- a/docs/newsletter/2025-12-memory-optimization.rst
+++ b/docs/newsletter/2025-12-memory-optimization.rst
@@ -39,6 +39,8 @@ To construct a multimodal mortality task in MIMIC-IV, we parse and scan:
      - 2.25
      - 3.97
 
+\*Results benchmarked on an AMD EPYC 7513 32-Core Processor Workstation.
+
 
 ----
 


dalloliogm:
Should I convert the VAE class to Pytorch Lightining? I understood other classes in PyHealth are based on that, and I like PL syntax better.

pyhealth/datasets/base_dataset.py
jhnwu3, line 1: Hey, we're just assuming absolute paths here.

jhnwu3: I guess that might be worth adding to the docs.

pyhealth/models/vae.py
```diff
@@ -26,129 +26,245 @@ class VAE(BaseModel):
...
+        input_size: input image size (for images, e.g. 128). Required if input_type='image'.
+        hidden_dim: the latent dimension. Default is 128.
+        conditional_feature_keys: list of feature keys to use as conditions for generation (optional).
+        **kwargs: other parameters.
     """
 
     def __init__(
         self,
-        dataset: BaseSignalDataset,
+        dataset,
```
jhnwu3: The dataset should be dataset : SampleDataset here.

pyhealth/models/vae.py
```diff
@@ -26,129 +26,245 @@ class VAE(BaseModel):
...
         feature_keys: List[str],
         label_key: str,
-        input_channel: int,
-        input_size: int,
         mode: str,
+        input_type: str = "image",
```
jhnwu3: We might be able to make some assumptions based on the feature keys' types here based on the sample_dataset.output_schema and input_schema instead of using an input_type argument here. 


See:
https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/datasets/sample_dataset.py

Basically, we can check if an input and output is an "image" here or using an ImageProcessor or 

sequence/timeseries processors here.

pyhealth/models/vae.py
```diff
@@ -26,129 +26,245 @@ class VAE(BaseModel):
...
+        self.cond_proj: Optional[nn.Linear] = None   # for conditional metadata → latent
+        self.ts_proj: Optional[nn.Linear] = None     # for timeseries concatenated features → hidden_dim
+
+        if input_type == "image":
+            assert input_channel is not None and input_size is not None, \
+                "For image mode, input_channel and input_size must be provided"
+
+            # Embedding model for conditional features only (if used)
+            if conditional_feature_keys:
+                self.embedding_model = EmbeddingModel(dataset, embedding_dim=hidden_dim)
```
jhnwu3: Probably not needed for images in our case here.

pyhealth/models/vae.py
```diff
@@ -26,129 +26,245 @@ class VAE(BaseModel):
...
-        super(VAE, self).__init__(
-            dataset=dataset,
-            feature_keys=feature_keys,
-            label_key=label_key,
-            mode=mode,
-        )
+        super(VAE, self).__init__(dataset=dataset)
+        self.input_type = input_type
         self.hidden_dim = hidden_dim
+        self.conditional_feature_keys = conditional_feature_keys
```
jhnwu3: For conditional features, maybe we can simplify our assumption to the basic multiclass representation/assumption here where we have a specific class we're generating for.

jhnwu3 (changes requested):
I think another nice thing to test out here is if our TaskClass() can support a COVIDCXR19ImageGeneration Class here

Task details here:
https://colab.research.google.com/drive/1kKkkBVS_GclHoYTbnOtjyYnSee79hsyT?usp=sharing


We could probably assume a labelprocessor or multihot schema for an input and its image. This way everything is force-defined. 

We can also throw an error if its some modality the VAE doesn't support currently.

Note that there's many ways to do a VAE, with different methods of tokenization, embeddings, etc. (Encoder-Decoder models, etc.) 

We can probably stick with our simple assumption here surrounding just image-based variables.

Discrete sequence generation would be a little more complicated (i.e autoregression) https://openreview.net/pdf?id=fYerSwf1Tb

dalloliogm:
Hi, I've accidentally closed this PR while trying to sync it with the release upstream. I may open a new one directly.

PR #654, opened 2025-12-05
[CS598] Add Mamba model support, new variants and updated tasks, and LR scheduler support for Trainer

kwand:
Daniel Kwan (NetID: dwkwan2)

This PR implements the following changes:

1. Preliminary support for the Mamba model, loosely based on reproduction of EHRMamba (https://arxiv.org/abs/2405.14567). Depends on the pure PyTorch mamba.py implementation of Mamba (https://github.com/alxndrTL/mamba.py, i.e. `pip install mambapy` - a library which is also used in HF Transformers) for building the actual Mamba blocks, which itself optionally depends on the official `mamba_ssm` implementation for their CUDA implementation (`use_cuda` currently set to false, to minimize dependencies) 
2. Two new variants of existing clinical predictive tasks (as specified in EHRMamba) on the MIMIC-IV dataset:
    - Mortality prediction within 31 days
    - Binary length-of-stay prediction (greater than one week or not)
3. Provides an updated version of the existing multi-class length-of-stay prediction function into the newer PyHealth task format (as the old function no longer works)
4. Implement LR scheduler support to the PyHealth Trainer, which can take in a LRScheduler class or optionally, a function that builds a scheduler (such as a SequentialLR for chaining together linear warmup and decay)

### DIFF
--- a/examples/example_mamba.ipynb
+++ b/examples/example_mamba.ipynb
@@ -1,0 +1,512 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "eb8b8ad6",
+   "metadata": {},
+   "source": [
+    "# Mamba on MIMIC-IV (Demo)\n",
+    "This minimal walkthrough uses PyHealth's built-in utilities to train the PyHealth Mamba model on a two new MIMIC-IV tasks: mortality prediction within one month, and binary LOS prediction over one week. It also includes a demonstration of using LR schedulers with PyHealth's Trainer."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "f372eb74",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pyhealth.datasets import MIMIC4Dataset, get_dataloader, split_by_sample\n",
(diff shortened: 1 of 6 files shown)


kwand:
Hi John,

I added the requested test cases and an `example_mamba.ipynb` notebook to `examples/` for a minimal example on using the Mamba model - and also pushed some remaining code that I forgot to earlier.

I couldn't figure out whether tasks should also have unit tests under `tests/`, since there doesn't seem to be any currently for MIMIC-IV tasks aside from under the `todo/` folder, so I simply demonstrated the new tasks under the same notebook 

The notebook also demonstrates the small (and very niche) use-case of providing a builder fn for a LR scheduler - in most cases, you would simply provide the class as with the existing optimizer behaviour.

Thanks,

joshuasteier:
Hi @kwand, following up on John's Dec 19 review. Just needs a few more things before it can merge:

- [ ] Add install requirement to pyproject.toml or raise error if mambapy not installed
- [ ] Add docs/api/models/pyhealth.models.Mamba.rst
- [ ] Add example initialization to docstrings
- [ ] Resolve merge conflict in tasks/__init__.py

. Let me know and I'm happy to do a full review once these are addressed.

jhnwu3:
We can probably close this one @joshuasteier since I believe we have Mamba inside the repo now.

PR #680, opened 2025-12-07
[CS 598] Add LSTM Model

bmalis2:
**Type:** Model Contribution
**Net IDs:** bmalis2, divyesh3
**Paper:** Addressing Wearable Sleep Tracking Inequity: A New Dataset and Novel Methods for a Population with Sleep Disorders
**Paper Link:** https://proceedings.mlr.press/v248/wang24a.html

**Description**: Implements a LSTM model class and a helper function test_lstm to test the class. 

**Model**: pyhealth/models/lstm.py
**Test**: function test_lstm

Model can be tested with `python pyhealth/models/lstm.py`

### DIFF
--- a/pyhealth/models/lstm.py
+++ b/pyhealth/models/lstm.py
@@ -1,0 +1,90 @@
+import torch
+from torch import nn
+from pyhealth.datasets import SampleDataset
+from pyhealth.models import BaseModel
+"""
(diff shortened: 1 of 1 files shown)


Logiquo:
LSTM is already included in the RNN

PR #687, opened 2025-12-07
Memory Efficient Dataset

Logiquo:


pyhealth/processors/image_processor.py
```diff
@@ -98,3 +101,8 @@ def __repr__(self) -> str:
             f"to_tensor={self.to_tensor}, normalize={self.normalize}, "
             f"mean={self.mean}, std={self.std}, mode={self.mode})"
         )
+
+
+def _convert_mode(img: Image.Image, mode: str) -> Image.Image:
+    """Convert a PIL image to the requested mode."""
+    return img.convert(mode)
```
Logiquo, line 108: to avoid pickle issue for multi-process

pyproject.toml
```diff
@@ -35,10 +35,11 @@ dependencies = [
   "urllib3~=2.5.0",
   "numpy~=1.26.4",
   "tqdm",
-  "polars~=1.31.0",
+  "polars~=1.35.2",
```
Logiquo, line 38: IMPORTANT! older version has memory bug on join operation

pyhealth/datasets/sample_dataset.py
jhnwu3, line 1: Just some thoughts: 

- If possible, it would be nice to keep the API the same, it would be good to combine InMemoryDataset with just SampleDataset, and maybe just have an in_memory=True/false init argument here. 
- I also wonder if we can keep the legacy samples= argument here someway? Not sure if that's possible since so many unit tests require some cute initialization of pseudo samples as an example.
- It might look really ugly, but I feel like create_sample_dataset() is a nice wrapper that could be instead just the SampleDataset's init() call here. (I could be wrong here, but it would be nice to consolidate)
- If the in-memory sampledataset is too heavy and a trivial speedup, we can simply just maintain the existing SampleDataset here instead maybe. 
- I see a SampleBuilder as well, not sure if it's possible to use that within SampleDataset instead??

Logiquo:
Closed in favour of #712

PR #694, opened 2025-12-07
[CS 598] Add Bonn EEG dataset/task

ravi14il:
**Contribution Type:** Dataset and Task
**Net ID:** ravi14

**Description**: Adding the Bonn EEG dataset  that was used in the paper I was reproducing (Known as Epilepsy in the paper). This is used to validate the model's robustness in domain transfer when pretrained on Sleep EEG (SleepEDF) Dataset which is already in Pyhealth. 

This dataset was also a test of trying to use the BaseDataset to load raw signal data that is not associated with patients/events along with streaming the data during the task.
 

**Paper**: Multi-View Contrastive Learning for Robust Domain Adaptation in Medical Time Series Analysis
**Paper Link**: https://arxiv.org/abs/2506.22393
**Dataset Paper Link**:  https://www.upf.edu/documents/229517819/232450661/Andrzejak-PhysicalReviewE2001.pdf/0e9a54b8-8993-b400-743e-4d64fa29fb63

**Dataset**: BonnEEGDataset
**Task**: BonnEEGSeizureDetection 
**Test**: TestBonnEEGDataset

### DIFF
--- a/pyhealth/datasets/__init__.py
+++ b/pyhealth/datasets/__init__.py
@@ -84,3 +84,4 @@ from .utils import (
     load_processors,
     save_processors,
 )
+from .bonn_eeg import BonnEEGDataset
--- a/pyhealth/datasets/bonn_eeg.py
+++ b/pyhealth/datasets/bonn_eeg.py
@@ -1,0 +1,113 @@
+"""
+Name: Vignesh Ravi
+Net ID: ravi14
+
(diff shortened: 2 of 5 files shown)


joshuasteier:
Thanks for the contribution! Going to close this - Bonn EEG is a classic dataset but at 500 segments with no patient/event structure, it doesn't benefit from PyHealth's pipeline. We already have SleepEDF/SHHS/ISRUC for EEG. Interesting note about using it for domain transfer validation though.

PR #707, opened 2025-12-08
[BMS] StageNet with Attention

Logiquo:

**Contribution Type**: Model

**Description**
Create a new model *StageNet with Attention*, with MHA layers inserted between SA-LSTM and SA-CNN.

**Files to Review**
pyhealth/models/stagenet_mha.py
tests/core/test_stagenet_mha.py

### DIFF
--- a/pyhealth/models/stagenet_mha.py
+++ b/pyhealth/models/stagenet_mha.py
@@ -12,11 +12,11 @@ from .embedding import EmbeddingModel
 
 
 class StageNetAttentionLayer(nn.Module):
-    """StageNet layer.
+    """StageNetAttention layer.
 
     Paper: Stagenet: Stage-aware neural networks for health risk prediction. WWW 2020.
 
-    This layer is used in the StageNet model. But it can also be used as a
+    This layer is used in the StageAttentionNet model. But it can also be used as a
     standalone layer.
 
     Args:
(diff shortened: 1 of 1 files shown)


pyhealth/models/stagenet_mha.py
```diff
@@ -0,0 +1,703 @@
...
+import torch.nn as nn
+
+from pyhealth.datasets import SampleDataset
+from pyhealth.models import BaseModel
+from pyhealth.models.utils import get_last_visit
+
+from .embedding import EmbeddingModel
+
+
+class StageNetLayer(nn.Module):
```
jhnwu3: We should probably rename this to StageNetAttentionLayer to make it clear we're doing attention-based stagenet.

pyhealth/models/stagenet_mha.py
```diff
@@ -0,0 +1,703 @@
...
+        distance = torch.stack(distance)
+
+        key_padding_mask = None
+        if mask is not None:
+            key_padding_mask = (mask == 0).to(device=device, dtype=torch.bool)
+
+        attn_output, _ = self.mha(
+            hidden_seq, hidden_seq, hidden_seq, key_padding_mask=key_padding_mask
+        )
+        attn_output = self.attn_norm(attn_output + hidden_seq)
```
jhnwu3, line 266: Is it possible to hook the attention scores from self.mha here? This way, we can pull out the attention weights visualization + its gradients for both attention-grad visualization/ interpretability approach. It would be cool to see how AttentionGrad, DeepLift, and GIM compare 

Transformer hooks (ver. might not be required) https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/models/transformer.py 
and 
https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/interpret/methods/chefer.py

Chefer's attribution approach method.

The goal is to be able to visualize each feature's tokens contribution. We may need to rewrite later how we process the StageNet task to support more granular lab events support (i.e instead of a 10 vector, we give each lab event its own embedding).

jhnwu3 (changes requested):
Can we also add it to the docs/?

I think it makes sense to stack attention between each RNN pass. So it's probably good enough here. We just need to rename the StageNet layer and also check that we can pull out both the attention scores and its gradients for the attn-grad visualizations.

PR #715, opened 2025-12-08
Adding example use case of mimic4

kimpossible6:
1. uiuc-598 dlh group project - ziweim2, yilin17 
2. What type of contribution you are making: example use case
3. demonstrates how to load the MIMIC-IV dataset, apply the
`Readmission30DaysMIMIC4` task, split the sample dataset by patients,
and inspect a batch of features
4. replace the dir path and run


jhnwu3:
redundant, closed.

PR #721, opened 2025-12-08
Fairlearn Metrics on MIMIC-III

Fahad-Z:
Fairness metric testing of PyHealth RNN readmission prediction model using the Synthetic MIMIC-III dataset. Synthetic sensitive attributes are used for baseline auditing purposes and to align with the goals of the original research of FAMEWS. Must install Fairlearn and  PyHealth before running.


jhnwu3:
Closed, will still get credit for course.

PR #722, opened 2025-12-08
Add reproduction of Zhang et al. (2024) Adaptive Learning HAR

ritwikbiswas:


Logiquo:
Could you split the dataset & model into pyhealth corresponding directory, and add necessary tests?

PR #728, opened 2025-12-08
MedLink Bounty

Rian354:
PR for MedLink bounty

Tests:
To run the MedLink unit tests, from the project root run:

pytest tests/core/test_medlink.py (locally, 3 passed & 1 warning)

Model Implementation:

- Implemented the MedLink retrieval model on top of the current BaseModel / dataset API.
- Added unit tests with small synthetic data for MedLink.
- Added a Jupyter notebook that trains and evaluates MedLink on the MIMIC-III demo dataset.

Additions to "pyhealth/models/medlink/model.py":

- BaseModel-compatible "MedLink" class that takes a task-generated dataset (e.g., "SampleDataset" from "set_task") and "feature_keys".
- Vocabulary construction from the underlying task dataset using "dataset.get_all_tokens(...)" for queries and documents.
- Query and corpus encoders ("encode_queries", "encode_corpus") that produce sparse multi-hot representations.
- BM25-style scoring in "compute_scores", compatible with the IR-format data produced by the MedLink utilities.
- Combined retrieval and prediction loss in forward / get_loss, returning a scalar loss for training.

Other changes:

- Extended SampleDataset w/ get_all_tokens(key: str) to collect unique tokens across samples, used by MedLink for vocabulary building.
- Implemented BM25 and IR helpers in the pyhealth.models.medlink package:
  - BM25Okapi
  - convert_to_ir_format, tvt_split
  - generate_candidates, filter_by_candidates
  - get_bm25_hard_negatives, get_train_dataloader, get_eval_dataloader
- Exported MedLink via pyhealth.models.__init__, so users can do:  from pyhealth.models import MedLink

Added examples/medlink_mimic3.ipynb, a runnable notebook that:

Loads the MIMIC-III demo dataset via MIMIC3Dataset.

Defines a patient linkage task to generate query–candidate pairs.

Uses the MedLink helpers to build IR-format data and PyTorch dataloaders.

Trains and evaluates MedLink and reports ranking metrics.

Locally ran:

examples/medlink_mimic3.ipynb runs end-to-end on the MIMIC-III demo dataset.

The notebook includes a note on how to run the MedLink unit tests from project root.

Files to review:

pyhealth/datasets/sample_dataset.py – SampleDataset.get_all_tokens helper for vocabulary construction.

pyhealth/models/medlink/model.py – core MedLink model implementation.

pyhealth/models/medlink/bm25.py – BM25Okapi implementation used in the retrieval pipeline.

pyhealth/models/medlink/utils.py – IR-format conversion, TVT split, candidate generation, dataloaders.

pyhealth/models/__init__.py – export of MedLink.

tests/core/test_medlink.py – synthetic unit tests for MedLink (forward pass, encoders, score shapes).

examples/medlink_mimic3.ipynb – Jupyter notebook for training and evaluating MedLink on the MIMIC-III demo dataset.


examples/medlink_mimic3.ipynb
jhnwu3, line 1: Some quick thoughts that:

- Can we move the medlink task into the pyhealth.tasks module too? I actually think it'd be really helpful also to further have detailed documentation surrounding the query/document identifiers. It'd be good to link it up with the original paper's task of mapping records to a master known patient record.

jhnwu3: It would also be nice to have it in the docs/ as that'll actually be a pretty nice to have for anyone working on record linkage problems.

pyhealth/models/medlink/model.py
jhnwu3, line 1: Can we also try to see if we can't build new processors here to pass to the MedLink model. 

Actually, I think the sequence processors should have built-in vocabularies here. But, it would be nice to update the EmbeddingModel to better support things like initialized Glove vectors or just use randomly initailized embeddings for now. This way medlink can be better integrated with the rest of PyHealth, and I think it'd be a nice lesson in replicating the original implementation. (A lot of the techniques are pretty relevant to clinical predictive modeling, so I think it's a good learning exercise).


Example of a PR working with the processors instead of the previous old PyHealth tokenizer approach here: https://github.com/sunlabuiuc/PyHealth/pull/610/files

Glove vectors from the original implementation: https://github.com/zzachw/MedLink here.

rianfyi:
### Summary

Updated with content that completes the Medlink bounty via adding the processor native implementation as requested (integrated w/ Pyhealth 2.x processors), unit tests with synthetic data, and an end to end MIMIC-III demo. It also has optional pretrained embedding initialization (glove style vectors) to better match the original Medlink implementation.

### Changes

#### 1) MedLink model implementation (processor native)
- Refactored pyhealth/models/medlink/model.py to use SequenceProcessor vocab over the legacy tokenizer approach.
- Builds a shared vocabulary across query and document processors (conditions and d_conditions) and remaps indices so embeddings are consistent between both.
- Contains randomly initialized embeddings by default, and optional pretrained embedding initialization (glove style) when given.

#### 2) Pretrained embeddings support
- Updated EmbeddingModel to support loading pretrained embedding vectors from a text file and optionally freezing them.
- This allows for reproducing the original Medlink workflow (optional) while keeping default behavior as just random initialization.

#### 3) Unit tests
- Added/updated tests/core/test_medlink.py with synthetic data to validate:
  - forward pass returns a scalar loss
  - query/corpus encoding shapes are consistent w/ vocab
  - score matrix shape matches (num_queries, num_docs)
- Tests are lightweight and do not require MIMIC.

#### 4) End-to-end example notebook
- Added examples/medlink_mimic3.ipynb showing:
  - loading MIMIC-III demo data
  - building patient linkage samples
  - training Medlink for a few epochs
  - evaluating with ranking metrics

### Bounty requirement coverage from feedback

- Implement MedLink model (using PyHealth 2.x patterns; processor-driven)
- Add a unit test with pseudo data that runs a forward pass (tests/core/test_medlink.py)
- Add an example notebook demonstrating training on MIMIC-III (examples/medlink_mimic3.ipynb)

### How to run tests

From repo root:

```bash
pip install -e .
pytest -q tests/core/test_medlink.py
```

pyhealth/datasets/sample_dataset.py
jhnwu3, line 1: ah, there's a merge conflict error here.

pyhealth/models/medlink.py
```diff
@@ -0,0 +1,423 @@
...
+
+    It implements three losses:
+      - forward admission prediction (corpus -> queries)
+      - backward admission prediction (queries -> corpus)
+      - retrieval loss via TF-IDF-style matching
+    """
+
+    def __init__(
+        self,
+        dataset: SampleEHRDataset,
```
jhnwu3: Canw e update our assumptions from SampleEHRDataset to SampleDataset here? 

SampleEHRDataset is quite defunct.

jhnwu3 (changes requested):
Very close! Thanks for the hardwork!

pyhealth/models/medlink.py
```diff
@@ -0,0 +1,423 @@
...
+        # In case of tiny mismatches, truncate to the smaller dimension.
+        batch = min(logits.size(0), targets.size(0))
+        logits = logits[:batch]
+        targets = targets[:batch]
+        return self.criterion(logits, targets)
+
+
+class MedLink(BaseModel):
+    """
+    MedLink: de-identified patient record linkage model (KDD 2023).
```
jhnwu3: Can we improve our doc strings here with an example initialization and usage case here? The autodocs will help with users looking to use these things later.

rianfyi: Yes, included usage for this segment. I also went over i/o for the whole file/model

rianfyi:
The tests seem to initialize sampledataset w/ samples parameter, but its expecting a schema file. So I added a helper to build the dataset w/ in memory samples. Should pass tests now

tests/core/test_sdoh.py
```diff
@@ -1,5 +1,5 @@
 from typing import Set
-from base import BaseTestCase
+from tests.base import BaseTestCase
```
jhnwu3, line 2: I don't really understand this change? Is it just because of the Python versioning? Doesn't seem to lead to issues though.

pyhealth/models/medlink.py
```diff
@@ -0,0 +1,447 @@
...
+        - The model primarily outputs a dictionary {"loss": loss_tensor} during training (see forward method).
+        - For retrieval/evaluation, the model provides embeddings and search utilities to score record similarity.
+
+    Example:
+        >>> from pyhealth.datasets import SampleDataset
+        >>> from pyhealth.models import MedLink
+        >>> samples = [{"patient_id": "1", "admissions": ["ICD9_430", "ICD9_401"]}, ...]
+        >>> input_schema = {"admissions": "code"}
+        >>> output_schema = {"label": "binary"}
+        >>> dataset = SampleDataset(path="/some/path", samples=samples, input_schema=input_schema, output_schema=output_schema)
```
jhnwu3: The SampleDataset example here needs to use the create_sample_dataset() function

pyhealth/models/medlink.py
```diff
@@ -0,0 +1,447 @@
...
+        **kwargs,
+    ) -> None:
+        # MedLink is defined over a single textual / code sequence feature
+        assert len(feature_keys) == 1, "MedLink supports exactly one feature key"
+        # BaseModel only accepts dataset parameter, not feature_keys, label_key, or mode
+        super().__init__(dataset=dataset)
+        # Set feature_keys manually since BaseModel extracts it from dataset.input_schema
+        # but MedLink needs to use the provided feature_keys
+        self.feature_keys = feature_keys
+        self.feature_key = feature_keys[0]
```
jhnwu3: Why the first one specifically? If need be, we can always define a medlink processor to make certain assumptions on the inputs.

For reference on processors, see here: https://pyhealth.readthedocs.io/en/latest/api/processors.html

Essentially, a tokenizer is a type of processor in our framework.

pyhealth/models/medlink.py
```diff
@@ -0,0 +1,447 @@
...
+            feature_size=embedding_dim,
+            heads=heads,
+            dropout=dropout,
+            num_layers=num_layers,
+        )
+
+        self.criterion = nn.BCEWithLogitsLoss()
+
+    def _encode_tokens(self, seqs: List[List[str]], device: torch.device):
+        """
```
jhnwu3: For instance, this here could be a processor _call_()

Edit: I just realized the forward() call here doesn't use _encode_tokens?

pyhealth/tasks/patient_linkage_mimic3.py
```diff
@@ -0,0 +1,106 @@
...
+    Output sample schema:
+      - patient_id: ground-truth entity id (equivalent to "master patient record id" in MIMIC)
+      - visit_id: query admission id (hadm_id)
+      - conditions, age, identifiers: query side
+      - d_visit_id: doc admission id (hadm_id)
+      - d_conditions, d_age, d_identifiers: doc side
+    """
+
+    task_name = "patient_linkage_mimic3"
+    input_schema = {"conditions": "sequence", "d_conditions": "sequence"}
```
jhnwu3: If you construct a MedLink processor, we can also change our schemas to be something like "q_conditions" : "medlink" (you can also use the class name itself like "q_conditions" : MedLinkProcessor() if that's easier to map in code) and "d_conditions" : "medlink" if that helps since it seems like the model does expect a sequence, but it encodes it in a different way with different Glove embeddings rather than the current EmbeddingModel approach?

pyhealth/models/medlink.py
```diff
@@ -0,0 +1,447 @@
...
+        embedding_dim: int,
+        heads: int = 2,
+        dropout: float = 0.5,
+        num_layers: int = 1,
+    ) -> None:
+        super().__init__()
+        self.tokenizer = tokenizer
+        self.vocab_size = tokenizer.get_vocabulary_size()
+
+        self.embedding = nn.Embedding(
```
jhnwu3: I don't see how this links up with the EmbeddingModel changes to use Glove vectors?

pyhealth/tasks/patient_linkage_mimic3.py
jhnwu3: You can also make a List[str] processor if you want to assume strings or something of that sort. Right now, the schemas don't seem to mean much at all, and thus it's just code that adds to the confusion of what is being inputted and outputted here. 

For instance, the doc strings don't match up with the code here.

pyhealth/models/medlink.py
```diff
@@ -0,0 +1,447 @@
...
+
+    Outputs:
+        - The model primarily outputs a dictionary {"loss": loss_tensor} during training (see forward method).
+        - For retrieval/evaluation, the model provides embeddings and search utilities to score record similarity.
+
+    Example:
+        >>> from pyhealth.datasets import SampleDataset
+        >>> from pyhealth.models import MedLink
+        >>> samples = [{"patient_id": "1", "admissions": ["ICD9_430", "ICD9_401"]}, ...]
+        >>> input_schema = {"admissions": "code"}
```
jhnwu3: I think we can simply assume that only certain tasks are compatible with the model, so we don't have to explicitly specify a feature_keys argument here that makes things a little more confusing and more opaque to the user.

tests/core/test_sdoh.py
rianfyi: This uses an absolute import instead of an implicit relative import. It makes sure that BaseTestCase is correctly resolved regardless of which directory the test runner is started from. Locally this did solve a ModuleNotFoundError.

pyhealth/models/medlink.py
rianfyi: Fixed.

pyhealth/models/medlink.py
rianfyi: I've kept the single feature assertion for now to stay close to the original paper's design, but the logic for extracting tokens is now correlated to the dataset's processors as mentioned above. I've also added some more detailed error messages if the expected keys (e.g. d_conditions) are missing in the dataset.

pyhealth/models/medlink.py
rianfyi: It actually does, but indirectly. forward() calls enc.logits_and_targets(), and logits_and_targets() calls 
_encode_tokens(). I've made sure this call chain is preserved and functioning correctly in the refactored AdmissionPrediction class (now in medlink/model.py).

pyhealth/tasks/patient_linkage_mimic3.py
rianfyi: With the new architecture where MedLink consumes SequenceProcessor logic directly via dataset.input_processors, the schema {"conditions": "sequence"} is fully accurate. We don't need a special "MedLinkProcessor" because we use standard SequenceProcessor for tokenization and vocab building, only handling of the embedding initialization differently in the model. Which keeps the schema standard. I have also updated PatientLinkageMIMIC3Task.input_schema to explicitly list all output fields (age, identifiers, etc.), which addresses the mismatch between docstrings and code.

pyhealth/models/medlink.py
rianfyi: This has been addressed in the latest changes. MedLink now takes a pretrained_emb_path argument, which is passed down to AdmissionEncoder. Inside AdmissionEncoder, we explicitly check for this path and call 
init_embedding_with_pretrained(self.embedding, ...) if it exists. This connects the model's nn.Embedding layer directly to the GloVe loading logic you saw in EmbeddingModel.

pyhealth/models/medlink.py
rianfyi: Agreed. I have updated MedLink.__init__ to make feature_keys optional. If not provided, the model now attempts to infer the correct sequence feature key from the dataset.input_processors (looking for a 
SequenceProcessor key k that also has a corresponding d_k processor). This simplifies the API while holding flexibility if manual specification is needed.

rianfyi:
All relevant tests pass as well.

pyhealth/tasks/patient_linkage_mimic3.py
```diff
@@ -0,0 +1,113 @@
...
+
+class PatientLinkageMIMIC3Task(BaseTask):
+    """
+    Patient linkage task for MIMIC-III using the Patient/Visit/Event API.
+
+    Produces the same sample keys as the original patient_linkage_mimic3 task
+    so pyhealth.models.medlink.convert_to_ir_format works as usual
+
+    Output sample schema:
+      - patient_id: ground-truth entity id (equivalent to "master patient record id" in MIMIC)
```
jhnwu3: The doc string output sample schema is erroneously labeled here as the input_schema contains all of the conditions you're looking at, which are the correct inputs, but medlink doesn't really output those, but rather should return a record list later after training for reference. In some sense, the d_conditions are the records being retrieved (although they are inputs for the medlink model too). 

https://dl.acm.org/doi/10.1145/3580305.3599427

Looking at the MedLink paper ^

"It is important to note that in practice, a patient may have multiple records in the database. To address this, we concatenate the 𝑘 most recent records from the patient with a special separator token (e.g., "[SEP]") to form a master record for the patient." (Page 3-4)"

https://claude.ai/share/db9be6a8-7bbd-4426-af53-0ccf72a621b7 

Here's a claude conversation link discussing some of the specifics. Tldr; We need to make sure we use the rest of the database records, and concatenate them into one large set sequence here. You'll need to add a "SEP" token during processing here.

jhnwu3 (changes requested):
I've added another comment to the task, let me know what you think. I'll look more at the EmbeddingModel, and other changes when I get the chance later.

pyhealth/tasks/patient_linkage_mimic3.py
rianfyi: Fixed, changes below:

1. **Pair Construction**: The task now creates pairs where:
    query ('r_q'): The last admission.
    Database Record ('r_d'): A concatenation of **all** previous admissions (sorted chronologically).

2. **[SEP] Token**: Inserted '[SEP]' tokens between admissions in the concatenated database record.

3. **Metadata**: Added 'time_gap_days' (time between query and most recent DB record) and timestamps to the output sample, which aligns w/ the paper's time interval analysis.

4. **One Pair Per Patient**: The task now returns a single positive pair per patient, as negatives will be handled via in batch sampling & hard negative mining during the training loop.

I've verified this with a test case that mocks a patient with multiple admissions and confirms the 'd_conditions' list contains the concatenated codes separated by '[SEP]'. I have also confirmed all relevant tests pass.

demo_medlink_fix.py
jhnwu3, line 1: Can we delete this file? It shouldn't be here.

pyhealth/tasks/patient_linkage_mimic3.py
```diff
@@ -0,0 +1,140 @@
...
+from pyhealth.tasks import BaseTask
+
+class PatientLinkageMIMIC3Task(BaseTask):
+    """
+    Patient linkage task for MIMIC-III.
+
+    For each patient with >=2 admissions:
+    - Query: last admission
+    - Positive database record: all previous admissions concatenated
+
```
jhnwu3, line 13: Can we add an example here in the docstrings on how to use this?

examples/medlink_mimic3.ipynb
jhnwu3, line 1: Anyway to use the PyHealth trainer here? It feels like you can already return a loss function given a batch. I see you also have a normalize function, that could be a new medlink processor in processors/

examples/test_eICU_addition.py
jhnwu3, line 1: Can we remove the eICU stuff from your PR if it's just about medlink? Or am I misunderstanding the exmaple usage here.

pyhealth/tasks/patient_linkage_mimic3.py
jhnwu3, line 13: Can we add better doc strings here on showing how to use this task

pyhealth/models/medlink/model.py
```diff
@@ -1,197 +1,415 @@
...
-               self.beta * bwd_cls_loss + \
-               self.gamma * ret_loss
+        loss = self.alpha * fwd_cls_loss + self.beta * bwd_cls_loss + self.gamma * ret_loss
         return {"loss": loss}
 
+    # ------------------------------------------------------------------
+    # Retrieval API
+    # ------------------------------------------------------------------
+
     def search(self, queries_ids, queries_embeddings, corpus_ids, corpus_embeddings):
```
jhnwu3, line 420: Can we add better docstrings on what's happening here? Specifically, that it's returning the c_idx, with some similarity score metrics I'm guessing?

pyhealth/models/medlink/model.py
```diff
@@ -1,197 +1,415 @@
...
+        emb[:, self.code_vocab.get("<pad>", 0)] = 0
+        if "<cls>" in self.code_vocab:
+            emb[:, self.code_vocab["<cls>"]] = 0
+        return emb
+
+    # ------------------------------------------------------------------
+    # Scoring / losses
+    # ------------------------------------------------------------------
+
+    def compute_scores(self, queries_emb: torch.Tensor, corpus_emb: torch.Tensor) -> torch.Tensor:
```
jhnwu3, line 365: Can you explain what we're doing here?

jhnwu3 (changes requested):
I think the main thing is just updating the docs/ or docstrings for each of the functions and the main medlink class. There are couple other things that I think should be cleaned up. Like the random demo_medlink_fix.py file in root directory.

But, it would be good to define exactly what each step of the process is trying to do with each of the sub-functions for more clarity on exactly what's happening here in Medlink.

rianfyi:
I also added more comprehensive docstrings to all functions in pyhealth/models/medlink/utils.py (which were previously undocumented). Updated the smoke test in model.py to use the new PatientLinkageMIMIC3Task class. All tests (including the new one) pass. On another note, can we delete this comment above? I believe its some sort of crypto scam.

pyhealth/models/medlink/model.py
rianfyi: Added a docstring explaining the scoring mechanism. The method computes a similarity by taking the dot product of the query embeddings (weighted by inverse document frequency of the codes) and corpus embeddings.

pyhealth/models/medlink/model.py
rianfyi: Search docstring now shows the inputs and the return format. So now, it specifies that it returns the nested dictionary mapping each query_id to a dictionary of corpus_id: score pairs.

examples/test_eICU_addition.py
rianfyi: Fair enough, this was moreso to address some previous pytest errors, its deleted.

examples/medlink_mimic3.ipynb
rianfyi: The notebook uses the trainer right now for the primary training loop. For the normalize function, I have kept it in the example for clarity/detail on the data formatting requirements, but I agree it's a viable candidate for a dedicated processor in the future, I think we'd have to discuss generalization/usage throughout the repo a bit first, and maybe some testing.

pyhealth/tasks/patient_linkage_mimic3.py
rianfyi: Added a usage example docstring. It shows the example of how to initialize dataset with the task and shows the exact structure of the results.

demo_medlink_fix.py
rianfyi: Done, deleted demo_medlink_fix.py and examples/test_eICU_addition.py

bbberns:
## Sovereign OS: High-Resolution Optimization
Documentation improvement proposed for project stability. Payout to LTC: ltc1qx2g4z3uk2gzux77k8gkyd0aduzfx30rane4gds

PR #743, opened 2025-12-18
Add num_workers to BaseDataset

Logiquo:

Contribution Type: Dataset

Description
Support configure `n_workers` for dask cluster.

Files to Review
pyhealth/datasets/base_dataset.py

### DIFF
--- a/pyhealth/datasets/base_dataset.py
+++ b/pyhealth/datasets/base_dataset.py
@@ -74,6 +74,7 @@ def path_exists(path: str) -> bool:
     else:
         return Path(path).exists()
 
+
 def _csv_tsv_gz_path(path: str) -> str:
     """
     Get the path to the file, trying the original path first, then the alternative path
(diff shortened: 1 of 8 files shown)


pyhealth/datasets/base_dataset.py
```diff
@@ -366,18 +369,14 @@ def global_event_df(self) -> pl.LazyFrame:
...
                 # Use cache_dir for Dask's scratch space to avoid filling up /tmp or home directory
                 dask_scratch_dir = self.cache_dir / "dask_scratch"
                 dask_scratch_dir.mkdir(parents=True, exist_ok=True)
 
                 with LocalCluster(
-                    n_workers=1,
+                    n_workers=self.num_workers,
                     threads_per_worker=1,
-                    processes=False,
+                    processes=not in_notebook(),
```
jhnwu3, line 379: Just curious, why do we need to check if we're in a jupyter notebook or not?

Logiquo: jupyter notebook is a bit tricky when dealing with multiprocessing (due to the natural of Python). So we would typically prefer to not use it when possible.

e.g. main guard in the cell

PR #746, opened 2025-12-19
Fix incorrect null handling for patient_id and timestamp

Logiquo:

**Contribution Type**: Dataset

**Description**
Fix incorrect null handling for patient_id and timestamp columns, including composite timetsamps
Add test for them.


jhnwu3 (approved):
lgtm,

PR #320, opened 2025-04-08
Dev

jhnwu3:
Massive refactor:

This pull request includes several changes to improve the PyHealth framework, particularly in dataset management and configuration. The most important changes include the addition of a new hackathon announcement, updates to dataset initialization and configuration, and the introduction of a new base dataset class.

### Hackathon Announcement:
* [`hackthon/202410-sunlab-hackthon.md`](diffhunk://#diff-e8df31d069d737f495dd6c268d4129488c00f4e87c1cf72ba0f21c8bfcf9afe7R1-R48): Added a detailed announcement for the Sunlab PyHealth Hackathon 2024, including schedule, project ideas, and resources.

### Dataset Initialization:
* [`pyhealth/data/__init__.py`](diffhunk://#diff-c0489d9435ccb2bd027ebdc1117452c99932839685043df02dbdbb05188dd835L1-R1): Simplified import statements by removing unused imports.
* [`pyhealth/datasets/__init__.py`](diffhunk://#diff-7e6c3885e3c07fd7bee62aa436567e2eca261b4e4e1dea7038b2442f375a3447L1-R19): Commented out unused dataset imports and added new import for `BaseDataset`.

### New Base Dataset Class:
* [`pyhealth/datasets/base_dataset.py`](diffhunk://#diff-b59401f7c1ec49894c525c968595104fc34b9d9a26c3fbb99dadcc08034cb98dR1-R250): Introduced a new `BaseDataset` class with detailed methods for loading and processing dataset tables, handling joins, and generating task-specific sample datasets.

### Configuration Updates:
* [`pyhealth/datasets/configs/mimic3.yaml`](diffhunk://#diff-d10328643a72a95da064b76c6c25cf561b03b566567ac29a268f188313c87e76R1-R77): Added configuration for MIMIC-III dataset, including file paths, patient attributes, and table joins.
* [`pyhealth/datasets/configs/mimic4.yaml`](diffhunk://#diff-44486a17ac67c365fbf3db2b63af2a4894f4c8235bd275ed78dc213da85ff6deR1-R111): Added configuration for MIMIC-IV dataset, detailing file paths, patient attributes, and table joins.

### Miscellaneous Fixes:
[...]

PR #751, opened 2025-12-24
[PyHealth2.0 Bounty] MIMIC3 Readmission Prediction

EricSchrock:
## Overview
This PR updates the `readmission_prediction_mimic3_fn` for PyHealth 2.0. I'll complete the readmission prediction tasks for the other datasets once this PR merges. This early PR should make that next PR go more smoothly.

In addition to supporting PyHealth 2.0, the new `ReadmissionPredictionMIMIC3` task has the following improvements. First, it takes a `datetime.timedelta` instead of an `int` for the readmission window, making the input parameter clearer and more flexible. Second, it can optionally exclude visits where the patient was under 18 years old (a previously unimplemented `#todo`).

Besides the items listed above, `ReadmissionPredictionMIMIC3` should be logically equivalent to `readmission_prediction_mimic3_fn`.

## Testing
I added unit tests for `ReadmissionPredictionMIMIC3` and updated and ran `examples/readmission_mimic3_rnn.py`.

I also updated `examples/readmission_mimic3_fairness.py`, but it doesn't run as `fairness_metrics_fn` was disabled in #320.

## Additional Notes
Apologies for all the whitespace changes in `README.rst`. My editor automatically removes trailing whitespace and I had to update a code snippet in that file. Let me know if you'd rather not merge the whitespace changes and I can undo them.

I considered keeping `readmission_prediction_mimic3_fn` as a wrapper around `ReadmissionPredictionMIMIC3` for backwards compatibility (I saw #735 does this for the LoS tasks). However, I'm not sure the changes to the LoS tasks are actually backwards compatible. The new tasks return lists for conditions, procedures, and drugs. The old tasks used to return _**lists of lists**_ for these items but now as pass throughs they just return lists. I'm not sure if that change in behavior would impact users of the old tasks.

### DIFF
--- a/tests/core/test_mimic3_readmission_prediction.py
+++ b/tests/core/test_mimic3_readmission_prediction.py
@@ -38,6 +38,11 @@ class TestReadmissionPredictionMIMIC3(unittest.TestCase):
 
     @classmethod
     def tearDownClass(cls):
+        # Deleting the dataset cache tmp dir causes ResourceWarnings
+        # These are caused by unclosed files due to lazy loading and can be safely ignored in tests
+        import warnings
+        warnings.filterwarnings("ignore", category=ResourceWarning)
+
         cls.cache_dir0.cleanup()
         cls.cache_dir1.cleanup()
         cls.cache_dir2.cleanup()


tests/core/test_mimic3_readmission_prediction.py
jhnwu3, line 1: Ah, it might be easier to use the mimic3demo dataset, itself in test-resources/ . You may need to expand these test.csv files in this directory though with more patients/specific attributes, as they're a subset of the actual mimic3demo for speed purposes:
https://github.com/sunlabuiuc/PyHealth/tree/master/test-resources/core/mimic3demo

MIMIC3demo download:https://physionet.org/content/mimiciii-demo/1.4/

EricSchrock: @jhnwu3 Are you wanting to keep test-resources/core/mimic3demo as a subset of the real demo dataset or do you mind if I add fake entries?

If I create fake entries, it would allow me to precisely test corner cases, but I'm sure I could probably find existing entries to get 90% of the test coverage I'm currently getting, which is probably plenty.

jhnwu3: Sorry for the late reply, go for it. If you want to add more synthetic examples, that works for me.

test-resources/core/mimic3demo/ADMISSIONS.csv.gz
EricSchrock, line 1: The only change to this file was to change the admission and discharge times for `subject_id` 10088's first admission such that the admission was less than 18 years after the subject's date of birth.

jhnwu3 (approved):
lgtm! Thanks a lot!

PR #757, opened 2025-12-28
Fix MIMIC3 unit tests

EricSchrock:
While working on updating #751 to use `test-resources/core/mimic3demo`, I noticed that the MIMIC3 unit tests were being skipped because the path to `mimic3demo` was wrong. It looks like this has been broken since #607 so a few other issues have crept into the MIMIC3 unit tests that needed to be patched up.


Logiquo:
any particular reason we want to rename .csv.gz file to UPPER CASE? e.g. ADMISSIONS.csv.gz

test-resources/core/mimic3demo/ICUSTAYS.csv.gz
EricSchrock, line 1: This file needs to exist since `MIMIC3Dataset` always loads the `icustays` table. However, I only included the header row, as actual data isn't needed to pass the tests.

```csv
subject_id,intime,icustay_id,first_careunit,dbsource,last_careunit,outtime

```

EricSchrock:
> any particular reason we want to rename .csv.gz file to UPPER CASE? e.g. ADMISSIONS.csv.gz

@Logiquo `mimic3.yaml` defines the default file names expected. Currently, it is set to expect upper case.

Issue #763, opened 2026-01-01
Better docs for the cache behaviour.


Logiquo:
We want a better documentation in terms of the behaviour of the cache directory, otherwise the user may load incorrect cache if they change somthing that our code did not catch.

JairoTM-es:
Hi! I’d like to work on this issue if it’s still available.
I’m a new contributor and would be happy to improve the documentation for the cache behavior.

Logiquo:
> Hi! I’d like to work on this issue if it’s still available. I’m a new contributor and would be happy to improve the documentation for the cache behavior.

Hi @JairoTM-es , welcome to the project! We’d love to have your help on this. I’ve assigned the issue to you.

To get started, you might want to look at https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/datasets/base_dataset.py. If you have any questions about the specific cache logic while writing the docs, feel free to ask here. Looking forward to your PR!

rushammm:
Hi, I noticed this was assigned in January but there's no PR yet. Is this still being worked on? If not, I'd be happy to pick it up. No pressure to @JairoTM-es — just want to make sure work isn't duplicated.

jhnwu3:
Happy to take any help here testing/running this. Right now, we've had some small updates to reduce the chances of a race condition, but it's still imperfect. If you haven't already cached a dataset, and you run multiple async jobs, it will crash all of them sometimes as they all try to write to a new cache.

rushammm:
Thanks! I'll start by reading through base_dataset.py and working on the documentation side first. Might have some questions as I go, hope that's okay.

jhnwu3:
Sure, just a heads-up, it'll be easier for us to respond on Discord instead of github just due to the breadth of things going on right now.

rushammm:
Got it, thanks! Do you have a Discord invite link I can join?

jhnwu3:
Yup, see our site here for the discord link button: https://pyhealth.dev/ 

(I'm trying to drive traffic to this site instead haha)

rushammm:
Thanks, just joined! Also linked the PR (#997) to this issue.

PR #766, opened 2026-01-01
Update the default task cache path to include task parameter names and values

### LINKED ISSUE #764: `.set_task` gives write cache to the same directory for the same task with different configuration
This would require
1. each task provide a interface for a unique task name or task id per configuration for the task that takes parameters. 
2. change the BaseDataset to use this new task name than old `task.task_name`.

EricSchrock:
## Summary
This PR fixes #764 as well as a few items I missed in #751.

1. Update the default task cache path to include task parameter names and values. This allows you to use multiple sample datasets at a time as long as they use different task parameter values.
2. Document `ReadmissionPredictionMIMIC3` as a class instead of as a function.
3. Add missing calls to `SampleDataset.close()` in `test_mimic3_readmission_prediction.py`, `test_chestxray14.py`, and `test_caching.py`.
4. Use a `tempfile.TemporaryDirectory` to resolve `ResourceWarning`s for unclosed files in unit tests. Putting this in the `setUpClass` method means the temporary directory will be cleaned up automatically _**after**_ all cache files are closed.

## Example Task Cache Paths
**Before**
* `/tmp/tmpgtq6w6c3/tasks/ReadmissionPredictionMIMIC3/`

**After**
* `/tmp/tmpgtq6w6c3/tasks/ReadmissionPredictionMIMIC3_window=15-days-0-00-00_exclude_minors=True/`
* `/tmp/tmpgtq6w6c3/tasks/ReadmissionPredictionMIMIC3_window=5-days-0-00-00_exclude_minors=True/`
* `/tmp/tmpgtq6w6c3/tasks/ReadmissionPredictionMIMIC3_window=15-days-0-00-00_exclude_minors=False/`

**Before**
* `/tmp/tmp969wfup_/tasks/ChestXray14BinaryClassification/`

**After**
* `/tmp/tmp969wfup_/tasks/ChestXray14BinaryClassification_disease=cardiomegaly/`
* `/tmp/tmp969wfup_/tasks/ChestXray14BinaryClassification_disease=hernia/`

### DIFF
--- a/pyhealth/datasets/base_dataset.py
+++ b/pyhealth/datasets/base_dataset.py
@@ -815,8 +815,7 @@ class BaseDataset(ABC):
             task (Optional[BaseTask]): The task to set. Uses default task if None.
             num_workers (int): Number of workers for multi-threading. Default is `self.num_workers`.
             cache_dir (Optional[str]): Directory to cache samples after task transformation,
-                but without applying processors. Default is
-                {self.cache_dir}/tasks/{task_name}_{task_param1_name}={task_param1_value}_..._{task_paramn_name}={task_paramn_value}.
+                but without applying processors. Default is {self.cache_dir}/tasks/{task_name}_{uuid5(vars(task))}.
             cache_format (str): Deprecated. Only "parquet" is supported now.
             input_processors (Optional[Dict[str, FeatureProcessor]]):
                 Pre-fitted input processors. If provided, these will be used


pyhealth/datasets/base_dataset.py
```diff
@@ -846,8 +848,12 @@ def set_task(
             f"Setting task {task.task_name} for {self.dataset_name} base dataset..."
         )
 
+        task_cache_dir = task.task_name
+        for k, v in vars(task).items():
+            task_cache_dir += re.sub(r"[^A-Za-z0-9._=-]+", "-", f"_{k}={v}")
```
Logiquo: I would have a bit concern regarding the path length. certain platforms (like Winodws) has a restriction on the total length of the path in terms of chars, and if a task has lots of parameters, it could exceed it. Maybe convert this into a uuid-v5 to ensure the length can not explode?

EricSchrock: That makes sense. If we are going with a UUID v5 we can also consider adding more input to the UUID generation to make sure the task cache works as expects as people develop tasks. Would you see any concerns with including both `vars(task)` and `inspect.getsource(task.__class__)` in the hash input to the UUID generation? I'd include some fallback logic for cases where the class source is not available.

Logiquo: I think `inspect.getsource` is a bit overkill. a simple change of formatting may trigger a re-compute, and also has some concerns regarding the binary format (like .whl). 

I'd say `vars(task)` should be sufficent, we don't need to make it too complicated.

EricSchrock: I think I largely agree, but there are two different use cases in tension here. The first use case is someone playing with different parameter values for an existing task. In that case, `vars(task)` is sufficient and adding `inspect.getsource(task.__class__)` is overkill and more brittle (for the reasons you mentioned).

The second use case is someone developing their own tasks. I'd expect this to be a pretty common use case as I'd guess that tasks are the part of the PyHealth pipeline that people are most likely to need to customize. I know as I was developing `ReadmissionPredictionMIMIC3` in #751 I got stuck for a bit trying to figure out why my changes to the task weren't being reflected in the generated samples (because cached samples were being used). Perhaps better documentation (#763) and examples (e.g. in the unit tests) are enough to solve this? Perhaps there are other ways we can aid task development without making sample caching more brittle?

For this PR, I'll stick with `task.task_name` plus a UUID v5 generated from `vars(task)`. If we think of better ways to address my concerns around custom task development, that can be a follow on.

@JairoTM-es this discussion thread is helpful context for #763.

Logiquo (approved):

Logiquo:
LGTM, thanks!

Issue #765, opened 2026-01-01
Add a `.clear_cache` and `.clear_task_cache` method to avoid the need to manually delete the cache.


Logiquo:
`.clear_cache` should clear the dataset cache (including its corresponding task cache)
`.clear_task_cache` should only clear the related task cache based on (dataset, task) pair. 

We want to prefereably implement this after #764 to allow more accurate task cache deletion.

Eshaan-byte:
@Logiquo, can you assign me

Eshaan-byte:
PTAL @Logiquo

PR #770, opened 2026-01-04
Add clear_cache and clear_task_cache methods to BaseDataset

Eshaan-byte:
Implements #765 by adding two new methods to manage cache cleanup:

- clear_cache(): Clears entire dataset cache including global event dataframe and all task caches
- clear_task_cache(task=None): Clears only the specified task's cache while preserving global event cache and other task caches

Both methods handle non-existent caches gracefully and provide comprehensive logging.

### DIFF
--- a/pyhealth/datasets/base_dataset.py
+++ b/pyhealth/datasets/base_dataset.py
@@ -905,6 +905,78 @@ class BaseDataset(ABC):
             task_name=task.task_name,
         )
 
+    def clear_cache(self) -> None:
+        """Clears the entire dataset cache, including global event data and all task caches.
+
+        This method removes:
+        - The global event dataframe cache (global_event_df.parquet)
+        - All task-specific caches in the tasks/ directory
+        - Temporary files in the tmp/ directory
+
+        After calling this method, the dataset will need to be reprocessed from
+        scratch the next time it is accessed.
+
+        Note:
+            This operation cannot be undone. Use with caution.
+        """
+        cache_path = self.cache_dir
+
+        if cache_path.exists():
+            logger.info(f"Clearing entire dataset cache at {cache_path}")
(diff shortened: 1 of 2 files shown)


EricSchrock (changes requested):
Could you add tests for these new methods to `tests/core/test_caching.py`, both to prove it works as expected and to act as example code?

pyhealth/datasets/base_dataset.py
```diff
@@ -905,6 +905,78 @@ def set_task(
...
+            task = self.default_task
+
+        # Generate the same task cache directory name as in set_task
+        task_params = json.dumps(
+            vars(task),
+            sort_keys=True,
+            default=str
+        )
+        task_cache_dir = self.cache_dir / "tasks" / f"{task.task_name}_{uuid.uuid5(uuid.NAMESPACE_DNS, task_params)}"
+
```
Logiquo: Could you move this out a separate methods, and let both this method & set_task method call it? This ensures the consistency incase we want to change the path later.

pyhealth/datasets/base_dataset.py
```diff
@@ -905,6 +905,78 @@ def set_task(
...
+            task (Optional[BaseTask]): The task whose cache should be cleared.
+                If None, uses the default task.
+
+        Raises:
+            AssertionError: If no default task is found and task is None.
+
+        Note:
+            This operation cannot be undone. The task cache will need to be
+            regenerated the next time set_task is called with this task.
+        """
```
Logiquo, line 970: I think we would want to explicit to specify this is the clear the default cache dir of task in the docstring, given `set_task` can always override the default one by passing the additional args.

Logiquo:
CI has failed.

PR #787, opened 2026-01-15
[Conformal EEG] Fix TUEV dataset metadata permissions for shared cluster storage - main

lehendo:
Author: Arjun Chatterjee (arjunc4@illinois.edu)
Contribution Type: Bug Fix

Fix TUEV dataset metadata permissions for shared cluster storage.

Modified prepare_metadata() to write metadata CSVs to ~/.cache/pyhealth/tuev/
as fallback when data directory is not writable. Updated init() to check
both shared and cache locations for CSVs. This resolves PermissionError when
users lack write access to shared data directories on clusters.

Tested on SunLab cluster with both tuev_conventional_conformal.py and
tuev_covariate_shift_conformal.py - processed 370 patients (113,353 samples)
successfully without permission errors.

### DIFF
--- a/pyhealth/datasets/tuev.py
+++ b/pyhealth/datasets/tuev.py
@@ -81,6 +81,24 @@ class TUEVDataset(BaseDataset):
             raise ValueError("subset must be one of 'train', 'eval', or 'both'")
         
         self.prepare_metadata()
+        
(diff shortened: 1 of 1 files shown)


jhnwu3 (approved):
Thanks!

PR #790, opened 2026-01-18
ICD 9 V Code Prediction Task and Example

sharimk2:


jhnwu3 (changes requested):
Hey Sharim! Thanks for the help! Can we implement some unit tests sorta like how paul does it here?

https://github.com/sunlabuiuc/PyHealth/blob/master/tests/core/test_sdoh.py

Basically, it helps to just test that the class itself works. 

Other things would be adding to the docs/ (Cursor does a pretty good job here given some examples).

pyhealth/models/sdoh_icd9_task.txt
```diff
@@ -0,0 +1,158 @@
+# SDOH Medical Coding Task
```
jhnwu3: Anyway to refactor this as a *.py file or just have the prompt here in the llm.py file? 

I guess another thing we could do is have a sdoh_icd9_llm/ directory for your work here.

pyhealth/models/sdoh_icd9_llm.py
```diff
@@ -0,0 +1,152 @@
...
+PROMPT_PATH = os.path.join(os.path.dirname(__file__), "sdoh_icd9_task.txt")
+
+
+def _load_prompt_template() -> str:
+    with open(PROMPT_PATH, "r", encoding="utf-8") as f:
+        return f.read()
+
+
+class SDOHICD9LLM:
+    """Admission-level SDOH ICD-9 V-code detector using an LLM."""
```
jhnwu3: Can we add more doc strings on how a user might use this model here?

jhnwu3: Please see some examples: https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/models/stagenet.py

jhnwu3 (changes requested):

pyhealth/models/sdoh_icd9_llm.py
sharimk2: I added more doc strings, let me know what you think @jhnwu3

pyhealth/datasets/sdoh_icd9.py
```diff
@@ -0,0 +1,111 @@
...
+}
+
+
+class SDOHICD9NotesDataset:
+    """CSV-backed dataset for SDOH ICD-9 V-code detection from notes."""
+
+    def __init__(
+        self,
+        csv_path: str,
+        dataset_name: Optional[str] = None,
```
jhnwu3, line 30: Is this your own dataset or did you use MIMIC-III? Or do you have to download it from somewhere?

pyhealth/datasets/mimic3.py
```diff
@@ -85,3 +89,156 @@ def preprocess_noteevents(self, df: pl.LazyFrame) -> pl.LazyFrame:
             .alias("charttime")
         )
         return df
+
+
+class MIMIC3NoteDataset:
```
jhnwu3, line 94: I believe MIMIC3Dataset already supports the loading of notes iirc, so this feels a bit redundant.

pyhealth/tasks/sdoh_icd9_detection.py
```diff
@@ -0,0 +1,63 @@
...
+
+import torch
+
+from .base_task import BaseTask
+from .sdoh_utils import TARGET_CODES, codes_to_multihot
+
+
+class SDOHICD9AdmissionTask(BaseTask):
+    """Builds admission-level samples for SDOH ICD-9 V-code detection.
+
```
jhnwu3, line 11: Can we get docstrings on how to use this task. For reference, the .set_task() feature here that you shouldn't overwrite unless you really need to in each of the BaseDataset classes require you pass a task through a dataset.set_task(TaskClass) here as it will get parallelized and used by our backend here.

jhnwu3 (changes requested):
Hey Sharim,

I would recommend checking out the pyhealth.datasets and pyhealth.tasks tutorials here: https://pyhealth.readthedocs.io/en/latest/tutorials.html to better understand how to extend PyHealth with your own dataset.

I think if you have separate labels in your dataset that's stored somewhere else on top of MIMIC3, then we'd probably need to create a new Dataset(BaseDataset) class here, but otherwise if its just a MIMIC3 showcase, I would recommend using the existing APIs here.

PR #797, opened 2026-01-26
Add multimodal RNN support

jhnwu3:
This pull request introduces a new multimodal RNN model to the `pyhealth` library, enabling the handling of both sequential and non-sequential input features for clinical prediction tasks. It also provides a comprehensive example script for using the new model on the MIMIC-IV dataset for in-hospital mortality prediction. Additionally, the documentation and model API are updated to reflect these changes.

**Major changes include:**

### New Model: MultimodalRNN

* Added the `MultimodalRNN` class to `pyhealth.models.rnn`, which automatically distinguishes between sequential and non-sequential features and processes them appropriately (sequential features via RNN layers, non-sequential features via direct embedding and pooling). The model concatenates all feature representations for final prediction.
* Exported `MultimodalRNN` in the `pyhealth.models` package init file for public use.
* Updated the API documentation to include `MultimodalRNN`.

### Example Usage

* Added a detailed example script `mortality_mimic4_multimodal_rnn.py` demonstrating how to use the new `MultimodalRNN` model for mortality prediction with mixed feature types on the MIMIC-IV dataset. The script covers data loading, task setup, model training, evaluation, and sample predictions.

### Improvements and Bug Fixes

* Improved masking in RNN forward passes by using the absolute value before summing, ensuring correct mask computation even if embeddings sum to zero due to negative values.
* Refactored imports and type annotations in `pyhealth.models.rnn` to support the new model and feature classification logic.

### DIFF
--- a/pyhealth/models/rnn.py
+++ b/pyhealth/models/rnn.py
@@ -248,6 +248,7 @@ class RNN(BaseModel):
         for feature_key in self.feature_keys:
             x = embedded[feature_key]
             # Use abs() before sum to catch edge cases where embeddings sum to 0
+            # @TODO bug with 0 embedding sum can still persist if the embedding is all 0s but the mask is not all 0s. 
             # despite being valid values (e.g., [1.0, -1.0])
             mask = (torch.abs(x).sum(dim=-1) != 0).int()
             _, x = self.rnn[feature_key](x, mask)
@@ -447,6 +448,7 @@ class MultimodalRNN(BaseModel):
             x = embedded[feature_key]
             # Use abs() before sum to catch edge cases where embeddings sum to 0
             # despite being valid values (e.g., [1.0, -1.0])
+            # @TODO bug with 0 embedding sum can still persist if the embedding is all 0s but the mask is not all 0s. 
             mask = (torch.abs(x).sum(dim=-1) != 0).int()
             _, last_hidden = self.rnn[feature_key](x, mask)
             patient_emb.append(last_hidden)


pyhealth/models/rnn.py
```diff
@@ -235,7 +247,9 @@ def forward(self, **kwargs) -> Dict[str, torch.Tensor]:
         embedded = self.embedding_model(kwargs)
         for feature_key in self.feature_keys:
             x = embedded[feature_key]
-            mask = (x.sum(dim=-1) != 0).int()
+            # Use abs() before sum to catch edge cases where embeddings sum to 0
+            # despite being valid values (e.g., [1.0, -1.0])
+            mask = (torch.abs(x).sum(dim=-1) != 0).int()
```
Logiquo, line 253: I think [0.0, 0.0] could still be a valid embedding valid even for non-padding token, though very unlikely. But this can be considered as a temporary fix for now i think. 

We may want to add a TODO here.

jhnwu3: Yeah, I think we may need to change our EmbeddingModel to be able to return a pad/mask tensor itself through the use of our processor vocabulary, and have it as a used return option here.

jhnwu3: Because I don't think the mask here really does very much here lol the more I think about it. It just assumes that somehow a 0 embedding from the EmbeddingModel is properly assigning 0 values to the sequence embeddings, etc.

PR #805, opened 2026-01-31
Change hook-based non-linearity replacement to module-swap-based. 

Logiquo:

Contribution Type: Interpretability

Description
Originally the DeepLift & GIM requires the model to have hook to replace the activation function, which is both cumbersome and error-prone (e.g. the old stagenet_mha forget to add a hook for one of the softmax). 

The new method only requires that all model non-linearity is defined as `nn.Module`, and DeepLift and GIM will dynamtically swap them if needed. This makes both the model definination cleanear and fix a few small issues regarding GIM with stagenet attn.


Logiquo:
This has also been tested on the mortality prediction stagenet mimic4 interpretability experiment and gives the same numbers

jhnwu3 (approved):
lgtm

PR #809, opened 2026-02-03
EHRMamba Implementation Into PyHealth

lehendo:
Arjun Chatterjee
Model implementation

Adds EHRMamba, a Mamba model for electronic health records.
Closely modeled off of the paper - EHRMAMBA: Towards Generalizable and Scalable Foundation Models for Electronic Health Records

Main work:
 **pyhealth/models/ehrmamba.py**: New module with `RMSNorm`, `MambaBlock` (SSM with causal conv), and `EHRMamba` (`BaseModel`). Uses existing `EmbeddingModel`, per-feature Mamba stacks, and `get_last_visit` for the classification head. Compatible with current PyHealth tasks and Trainer.

Testing and example notebook will also be included. I also included an icustays csv file because the mimic4ehrdataset class has default table icustays which is not present anywhere in the codebase.

### DIFF
--- a/examples/conformal_eeg/tuev_kmeans_conformal.py
+++ b/examples/conformal_eeg/tuev_kmeans_conformal.py
@@ -1,0 +1,203 @@
+"""K-means Cluster-Based Conformal Prediction (ClusterLabel) on TUEV EEG Events using ContraWR.
+
+This script:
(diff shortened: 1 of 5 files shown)


jhnwu3 (approved):
lgtm

PR #811, opened 2026-02-04
Fix lime CUDA OOM

Logiquo:

Contribution Type: Interpretability

Description
The old implementation suffers from 
```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 20.00 MiB. GPU 5 has a total capacity of 47.53 GiB of which 17.06 MiB is free. Including non-PyTorch memory, this process has 47.51 GiB memory in use. Of the allocated memory 47.20 GiB is allocated by PyTorch, and 2.27 MiB is reserved by PyTorch but unallocated. 
```

The new code accumulate tensor in cpu while maintain most of the compute on gpu.

### DIFF
--- a/pyhealth/interpret/methods/lime.py
+++ b/pyhealth/interpret/methods/lime.py
@@ -410,33 +410,35 @@ class LimeExplainer(BaseInterpreter):
                 torch.ones(n_features, device=device) * 0.5
             )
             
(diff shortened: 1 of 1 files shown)


jhnwu3 (approved):
Lgtm

PR #821, opened 2026-02-06
Standard `forward_from_embedding` in BaseModel

### LINKED ISSUE #818: standard embedding BaseModel
standardize the forward_from_embedding method in base model

Logiquo:
Add three new methods `forward`, `forward_from_embedding`, `get_embedding_model`. 

This breaks the old `forward_from_embedding`, but make it more mimic towards `forward`. This should make the API a bit cleaner to avoid a bunch of random named args, but does mean all interpertability methods need to update according to new API

Fix #818.

# Models
- [x] MLP
- [x] StageNet
- [x] StageAttn
- [x] Transfromer

# Interpretability
- [x] Random
- [x] Lime (But get better performance)
- [x] Shap
- [x] Integrated Gradient
- [x] GIM
- [x] DeepLift
- [x] IG-GIM
- [ ] ~~Chefer~~ (Deferred to next PR due to coupling with attention)

### DIFF
--- a/pyhealth/interpret/methods/deeplift.py
+++ b/pyhealth/interpret/methods/deeplift.py
@@ -391,7 +391,19 @@ class DeepLift(BaseInterpreter):
             if "mask" in schema:
                 masks[k] = v[schema.index("mask")]
             else:
-                masks[k] = (v[schema.index("value")] != 0).int()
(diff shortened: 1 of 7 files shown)


jhnwu3 (approved):
lgtm. We can iterate if need be.

Issue #789, opened 2026-01-17
The `core.test_gamenet.TestGAMENet.test_loss_is_finite` unit test is unstable


EricSchrock:
While developing #788, I saw the `core.test_gamenet.TestGAMENet.test_loss_is_finite` unit test fail during a GitHub actions run, though I had not changed any related files. I reran the pipeline without changes, and it passed. Opening an issue ticket for awareness, in case others are intermittently seeing this issue too.

```sh
======================================================================
FAIL: test_loss_is_finite (core.test_gamenet.TestGAMENet.test_loss_is_finite)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/PyHealth/PyHealth/tests/core/test_gamenet.py", line 111, in test_loss_is_finite
    self.assertTrue(torch.isfinite(ret["loss"]).all())
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: tensor(False) is not true
```

https://github.com/EricSchrock/PyHealth/actions/runs/21098985432/attempts/1

lehendo:
Fixed https://github.com/sunlabuiuc/PyHealth/pull/824/changes/11b9eaa4bf3ebde84cc7ac7971695fc69fd9f9be

PR #824, opened 2026-02-06
[Conformal EEG] Neighborhood Conformal Prediction

lehendo:
Arjun Chatterjee

This an implementation of the neighborhood conformal prediction from the paper that Sid shared. Added a test file and changed some of the init files.

Main implementation:
Calibrate: Stores calibration embeddings and conformity scores (prob of true class). Fits a k-NN index on calibration embeddings. Optional cal_embeddings or batch_size for extraction.
Forward: One model call with embed=True. For each test embedding: k-NN in calibration space, weights = exp(-dist/λ_L) over k-NN and normalized, threshold = weighted α-quantile of calibration conformity scores, y_predset = (y_prob >= threshold) with correct batch handling.
Args: model, alpha, k_neighbors=50, lambda_L=100.0, debug=False.
Validation: alpha in (0,1), k_neighbors positive int, multiclass-only.

Fix https://github.com/sunlabuiuc/PyHealth/issues/789

### DIFF
--- a/examples/conformal_eeg/tuev_ncp_conformal.py
+++ b/examples/conformal_eeg/tuev_ncp_conformal.py
@@ -7,9 +7,14 @@ This script:
 4) Extracts calibration embeddings and calibrates a NeighborhoodLabel (NCP) predictor.
 5) Evaluates prediction-set coverage/miscoverage and efficiency on the test split.
 
+With --n-seeds > 1: fixes the test set (--split-seed), runs multiple training runs
+with different seeds (different train/val/cal splits and model init), reports
+coverage / set size / accuracy as mean ± std (error bars).
+
 Example (from repo root):
   python examples/conformal_eeg/tuev_ncp_conformal.py --root /srv/local/data/TUH/tuh_eeg_events/v2.0.0/edf
   python examples/conformal_eeg/tuev_ncp_conformal.py --quick-test --log-file quicktest_ncp.log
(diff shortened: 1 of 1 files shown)


tests/core/test_gamenet.py
EricSchrock, line 1: Fixes #789

pyhealth/calib/predictionset/cluster/neighborhood_label.py
```diff
@@ -0,0 +1,206 @@
...
+                "embeddings. Ensure it supports embed=True in forward()."
+            )
+
+        test_emb = pred["embed"].detach().cpu().numpy()
+        test_emb = np.atleast_2d(test_emb)
+        batch_size = test_emb.shape[0]
+        n_cal = self.cal_conformity_scores_.shape[0]
+        k = min(self.k_neighbors, n_cal)
+
+        distances, indices = self._nn.kneighbors(test_emb, n_neighbors=k)
```
siddharthal, line 211: This implementation for NCP is most likely wrong. 
check equations 2 in the paper: https://arxiv.org/pdf/2303.10694

The calibration is broken into two steps: 1) a q_ncp function, this calibrates local neighborhood of each calibration point. 2) write an alpha^tilde_NCP function for computing the ncp quantile.  3) check the distance functions defined in equations 3-5, we might want to check exponential transformation of the distance.

It appears this is what the current code does - identify K nearest neighbors of each test point; set importance weights based on distances from test point to these calibration points. This does not yield marginal coverage guarentees.

siddharthal (changes requested):

siddharthal (approved):
The updates on the NCP implementations look good to me!

PR #827, opened 2026-02-07
Add a few more methods to indicate the output data of a processor

Logiquo:
This propose a new set of methods need to be implemented for processor.

This should greatly reduce the complexity for both models & interpretability methods to generalize across multiple inputs, as they now have a standard way to understand the data they are receiving. 

new APIs are
```python
    def is_token(self) -> bool:
        """Returns whether the output (in particular, the value tensor) of the processor 
        represents discrete token indices (True) or continuous values (False). This is used to 
        determine whether to apply token-based transformations (e.g. `nn.Embedding`) or 
        value-based augmentations (e.g. `nn.Linear`). 

        Returns:
            True if the output of the processor represents discrete token indices, False otherwise.
        """
        raise NotImplementedError("is_token method is not implemented for this processor.")
    
    def schema(self) -> tuple[str, ...]:
        """Returns the schema of the processed feature. For a processor that emits a single tensor,
        this should just return `["value"]`. For a processor that emits a tuple of tensors, 
        this should return a tuple of the same length as the tuple, with the semantic name of each tensor,
        such as `["time", "value"]`, `["value", "mask"]`, etc.
        
        Typical semantic names include:
            - "value": the main processed tensor output of the processor
            - "time": the time tensor output of the processor (mostly for StageNet)
            - "mask": the mask tensor output of the processor (if applicable)
        
        Returns:
            Tuple of semantic names corresponding to the output of the processor.
        """
        raise NotImplementedError("Schema method is not implemented for this processor.")
    
    def dim(self) -> tuple[int, ...]:
        """Number of dimensions (`Tensor.dim()`) for each output
        tensor, in the same order as the output tuple.

        Returns:
            Tuple of integers corresponding to the number of dimensions of each output tensor.
        """
        raise NotImplementedError("dim method is not implemented for this processor.")
    
    def spatial(self, i: int) -> tuple[bool, ...]:
        """Whether each dimension (axis) of the i-th output tensor is spatial (i.e. corresponds to a spatial 
        axis like time, height, width, etc.) or not. This is used to determine how to apply 
        augmentations and other transformations that should only be applied to spatial dimensions.
        
        E.g. for CNN or RNN features, this would help determine which dimensions to apply spatial augmentations to, 
        and which dimensions to treat as channels or features.
        
        Args:
            i: Index of the output tensor to check.
        
        Returns:
            Tuple of booleans corresponding to whether each axis of the i-th output tensor is spatial or not.
        """
        raise NotImplementedError("spatial method is not implemented for this processor.")
    
```

### DIFF
--- a/pyhealth/processors/audio_processor.py
+++ b/pyhealth/processors/audio_processor.py
@@ -163,26 +163,16 @@ class AudioProcessor(FeatureProcessor):
             return (3,)
         return (2,)
 
-    def spatial(self, i: int) -> tuple[bool, ...]:
+    def spatial(self) -> tuple[bool, ...]:
         """Whether each dimension of the output tensor is spatial.
 
         For waveform (channels, samples): channels is not spatial, samples is.
         For mel spectrogram (channels, n_mels, time): channels is not spatial,
         n_mels and time are.
 
-        Args:
-            i: Index of the output tensor (must be 0).
-
         Returns:
             Tuple of booleans for each axis.
-
-        Raises:
-            IndexError: If i != 0 (only one output tensor).
         """
-        if i != 0:
-            raise IndexError(
-                f"AudioProcessor has 1 output tensor, but index {i} was requested."
-            )
         if self.n_mels is not None:
             # (channels, n_mels, time)
             return (False, True, True)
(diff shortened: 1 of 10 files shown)


Logiquo:
Closed in favour of #821

PR #835, opened 2026-02-09
init commit, tldr the cache_dir or files should probably indicate tab…

jhnwu3:
The problem here is it can be quite confusing for the user to know exactly what tables are cached inside each cache file when they want to specify a new location.

### DIFF
--- a/examples/benchmark_perf/benchmark_workers_1.py
+++ b/examples/benchmark_perf/benchmark_workers_1.py
@@ -117,6 +117,7 @@ def main():
     print("\n[1/2] Loading MIMIC-IV base dataset...")
     dataset_start = time.time()
 
+    base_cache_dir = f"{cache_root}/base_dataset"
     base_dataset = MIMIC4Dataset(
         ehr_root="/srv/local/data/physionet.org/files/mimiciv/2.2/",
         ehr_tables=[
@@ -127,7 +128,7 @@ def main():
             "labevents",
         ],
         dev=dev,
-        cache_dir=f"{cache_root}/base_dataset",
+        cache_dir=base_cache_dir,
     )
 
     dataset_time = time.time() - dataset_start
@@ -147,12 +148,10 @@ def main():
 
     # Measure cache sizes
     print("\n[3/3] Measuring cache sizes...")
-    base_cache_dir = f"{cache_root}/base_dataset"
-    task_cache_dir = f"{base_cache_dir}/tasks"
 
-    base_cache_size = get_directory_size(base_cache_dir)
-    task_cache_size = get_directory_size(task_cache_dir)
-    total_cache_size = base_cache_size + task_cache_size
+    total_cache_size = get_directory_size(base_cache_dir)
+    task_cache_size = get_directory_size(f"{base_cache_dir}/tasks")
+    base_cache_size = total_cache_size - task_cache_size
 
     print(f"✓ Base dataset cache: {format_size(base_cache_size)}")
     print(f"✓ Task samples cache: {format_size(task_cache_size)}")
--- a/examples/benchmark_perf/benchmark_workers_12.py
+++ b/examples/benchmark_perf/benchmark_workers_12.py
@@ -117,6 +117,7 @@ def main():
     print("\n[1/2] Loading MIMIC-IV base dataset...")
     dataset_start = time.time()
 
+    base_cache_dir = f"{cache_root}/base_dataset"
     base_dataset = MIMIC4Dataset(
         ehr_root="/srv/local/data/physionet.org/files/mimiciv/2.2/",
         ehr_tables=[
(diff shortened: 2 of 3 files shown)


jhnwu3:
Don't know how to change this to a draft. But, will keep iterating on this as I move along.

pyhealth/datasets/base_dataset.py
```diff
@@ -333,16 +346,59 @@ def __init__(
...
+        # at init time -- it's used as-is with no modification.
+        if isinstance(self._cache_dir, Path):
+            return self._cache_dir
+
         if self._cache_dir is None:
+            # No cache_dir provided: use default pyhealth cache directory
+            cache_dir = Path(platformdirs.user_cache_dir(appname="pyhealth")) / "datasets"
+            cache_dir.mkdir(parents=True, exist_ok=True)
+            logger.info(f"No cache_dir provided. Using default cache dir: {cache_dir}")
+            self._cache_dir = cache_dir
```
Logiquo: Because all task files are still in the same `.../tasks/{task.task_name}_{uuid.uuid5(uuid.NAMESPACE_DNS, task_params)}/...`

For the same task with same schema, same processor, but different dataset, it will still collide.

EricSchrock:
@jhnwu3 and @Logiquo I believe the updates I pushed fix both concerns while passing existing tests. I also added new tests for the new caching behavior. Let me know what you think.

1. Datasets with different configurations (root, tables, name, or dev value) get different caches.
2. Tasks with different configurations (names, schemas, or arg values) get different caches. Also, since default task caches are nested in dataset caches, running the same task on datasets with different configurations will result in different task caches, **unless you override the default task cache path**.

Maybe we feed the dataset UUID into the task UUID generation so that if someone overrides the default task path, task caching will still be sensitive to dataset configuration?

Or maybe we should remove the ability for users to override the default task cache? Do they really need to be able to override both the dataset and task cache paths?

pyhealth/datasets/base_dataset.py
```diff
@@ -822,18 +844,17 @@ def set_task(
...
-            samples_{uuid}.ld/  # Final processed samples after applying processors
-                schema.pkl      # Saved SampleBuilder schema
-                *.bin           # Processed sample files
-            samples_{uuid}.ld/
-                ...
+            {task_name}_{task_uuid}/        # Cached data for specific task based on task name, schema, and args
+                task_df.ld/                 # Intermediate task dataframe based on schema
+                samples.ld/                 # Final processed samples after applying processors
+                    schema.pkl              # Saved SampleBuilder schema
+                    *.bin                   # Processed sample files
```
Logiquo: schema itself is insufficient to capture the uniqueness, as the processor instance could be different even though task schema is the same.

We would still need `samples_<sample_uuid>.ld` here.

Logiquo: I think the code behaviour is still correct, only the docs here is a bit confussing.

pyhealth/datasets/base_dataset.py
```diff
@@ -822,18 +844,17 @@ def set_task(
...
 
-            task_df.ld/ # Intermediate task dataframe after task transformation
-            samples_{uuid}.ld/  # Final processed samples after applying processors
-                schema.pkl      # Saved SampleBuilder schema
-                *.bin           # Processed sample files
-            samples_{uuid}.ld/
-                ...
+            {task_name}_{task_uuid}/        # Cached data for specific task based on task name, schema, and args
+                task_df.ld/                 # Intermediate task dataframe based on schema
+                samples.ld/                 # Final processed samples after applying processors
```
Logiquo:
```suggestion
                samples_<proc_uuid>.ld/                 # Final processed samples after applying processors
```

pyhealth/datasets/base_dataset.py
```diff
@@ -326,44 +334,55 @@ def __init__(
...
+
+        The cache structure within the directory is::
+
+            {dataset_uuid}/                         # Cache files for this dataset configuration
+                tmp/                                # Temporary files during processing
+                global_event_df.parquet/            # Cached global event dataframe
+                tasks/                              # Cached task-specific data
+                    {task_name}_{task_uuid}/        # Cached data for specific task based on task name, schema, and args
+                        task_df.ld/                 # Intermediate task dataframe based on schema
+                        samples.ld/                 # Final processed samples after applying processors
```
Logiquo:
```suggestion
                        samples_<proc_uuid>.ld/                 # Final processed samples after applying processors
````

Logiquo: @EricSchrock Here is another one.

EricSchrock (changes requested):
I think the one thing left is to update the benchmarking examples. Without the ability to set the task cache, the math for calculating cache usage needs to be updated.

Logiquo (approved):
LGTM, thanks @jhnwu3 @EricSchrock

jhnwu3:
@EricSchrock I've updated benchmark scripts to use the correct cache_dir/tasks/ layout for calculating cache sizes. Let me know if this works.

PR #840, opened 2026-02-11
[Foundational-EHR-Model-Multimodal] Create initial task (discharge notes + radiology notes)

will-pang:
## Contributor Information
- Name: William Pang
- Contribution Type: Task
- Links: 
  - [Project Google Doc](https://docs.google.com/document/d/1mALf80aaZ5XzacFPy8Dwy1GnyqkfOIS1C0CniKbhszw/edit?tab=t.2elsrg4w3jmh)
  - [Tuple Time Processor](https://pyhealth.readthedocs.io/en/latest/api/processors/pyhealth.processors.TupleTimeTextProcessor.html)

#### v0.2
Per John's [feedback](https://github.com/sunlabuiuc/PyHealth/pull/840#issuecomment-3905246278) I've incorporated a few changes:
(1) Updated `_compute_time_diffs(notes_with_timestamps, first_admission_time)` helper function
 - Sorts notes chronologically by timestamp, then computes each note's offset (in hours) from the first admission time. 
   - If a member does not have a note, it will return `(["<missing>"], [0.0])`
 - TODO: I should do a bit more testing on `_compute_time_diffs` just to see if there are edge-cases where (1) `first_admission_time` is missing (2) `first_admission_time` does not make sense relative to the timestamps of other notes. I should probably connect with @jhnwu3 to understand if there are edge cases to look out for.
 
(2) Added text tokenizer to @Rian354's `pyhealth/processors/tuple_time_text_processor.py` script
  - In `pyhealth/tasks/ehr_foundational_model_mimic4.py` you can also now pass the tokenizer settings in `kwargs` (see example #[5](https://pyhealth.readthedocs.io/en/latest/api/processors.html#) from processors docs)
  
**Testing Notes**
- To run the task pipeline, you can run this script: `examples/foundation_ehr/multimodal_task.py`
- To test the task, I've written a very simple unit test for now: `uv run python -m unittest tests.core.test_ehr_foundational_model_mimic4`
---

#### v0.1
Per John's [feedback](https://github.com/sunlabuiuc/PyHealth/pull/840#pullrequestreview-3782473464), I've incorporated a few changes:
- Incorporate Rian's `tuple_time_text_processor`, which now feeds in radiology and discharge notes as `(note_text, time_diff_hours)` tuples
  - I think we still need to discuss how to compute `time_diff`, in the sense that whether the timestamps from `note.timestamp` give a proper chronology of time or not. If we need to arrange in chronological order, I probably need to add a `.sort(lambda x: x['time_stamp'])` function or something equivalent.
- If the note is missing, we now return:
  ```
  return (["<missing>"], [0.0]) # TODO: How should we handle notes with missing timestamps? 
  ```
- The code seems to bug out on the preprocessor. I'm not as familiar with litdata for serialization,  but per Claude:

> When litdata sees a tuple like (List[str], List[float], str), it flattens it and infers a single serializer for the whole field based on the first element it encounters. So if the first element it hits is a str, it uses the string serializer for everything — then fails when it encounters a float.

---
#### v0
More of a draft PR as I'm still fairly new to the inner workings of the package, but here'a few things that I still think needs to be done:

- Seems to work when I run:
```
  dataset = MIMIC4Dataset(
          ehr_root=EHR_ROOT,
          note_root=NOTE_ROOT,
          ehr_tables=["diagnoses_icd", "procedures_icd", "prescriptions", "labevents"],
          note_tables=["discharge", "radiology"],
          cache_dir=CACHE_DIR,
          num_workers=16,
          dev = True
      )
  
      task = EHRFoundationalModelMIMIC4()    
      samples = dataset.set_task(task, cache_dir=f"{CACHE_DIR}/task", num_workers=8)
```
but when I run it outside of dev mode, I run into this error:

```
AttributeError: 'str' object has no attribute 'dtype'                                                             
/Users/wpang/.local/share/uv/python/cpython-3.12.12-macos-aarch64-none/lib/python3.12/multiprocessing/resource_tr 
acker.py:279: UserWarning: resource_tracker: There appear to be 10 leaked semaphore objects to clean up at        
shutdown                                                                                                          
warnings.warn('resource_tracker: There appear to be %d '      
```  

Claude says that this relates to the notes varying by length from patient to patient (e.g., patient A might have 4 radiology notes and 2 discharge notes, whereas patient B might have 2 radiology notes and 5 discharge notes), but I'm a little stuck as I am still getting comfortable with the architecture of the package.

**Testing Notes**
- To test, you can run this script: `examples/foundation_ehr/multimodal_task.py`
---

### DIFF
--- a/pyhealth/tasks/ehr_foundational_model_mimic4.py
+++ b/pyhealth/tasks/ehr_foundational_model_mimic4.py
@@ -25,12 +25,7 @@ class EHRFoundationalModelMIMIC4(BaseTask):
                     "tokenizer_name": "bert-base-uncased",
                     "type_tag": "note",
                 },
-            ),
-            "icd_codes": (
-                "stagenet", 
-                {"padding": 0
-                }
-            ),
+            )
         }
         self.output_schema: Dict[str, str] = {"mortality": "binary"}
 
@@ -109,8 +104,8 @@ class EHRFoundationalModelMIMIC4(BaseTask):
                 except AttributeError: # note object is missing .text or .timestamp attribute (e.g. malformed note)
                     pass
             if not discharge_notes: # If we get an empty list
-                all_discharge_texts.append(TOKEN_REPRESENTING_MISSING_TEXT) # Token representing missing text
-                all_discharge_times_from_admission.append(TOKEN_REPRESENTING_MISSING_FLOAT) # Token representing missing time(?)
+                all_discharge_texts.append(self.TOKEN_REPRESENTING_MISSING_TEXT) # Token representing missing text
+                all_discharge_times_from_admission.append(self.TOKEN_REPRESENTING_MISSING_FLOAT) # Token representing missing time(?)
 
             for note in radiology_notes: #TODO: Maybe make this into a helper function?
                 try:
@@ -124,14 +119,11 @@ class EHRFoundationalModelMIMIC4(BaseTask):
                 except AttributeError: # note object is missing .text or .timestamp attribute (e.g. malformed note)
                     pass
             if not radiology_notes: # If we receive empty list
-                all_radiology_texts.append(TOKEN_REPRESENTING_MISSING_TEXT) # Token representing missing text
-                all_radiology_times_from_admission.append(TOKEN_REPRESENTING_MISSING_FLOAT) # Token representing missing time(?)
+                all_radiology_texts.append(self.TOKEN_REPRESENTING_MISSING_TEXT) # Token representing missing text
+                all_radiology_times_from_admission.append(self.TOKEN_REPRESENTING_MISSING_FLOAT) # Token representing missing time(?)
 
         discharge_note_times_from_admission = (all_discharge_texts, all_discharge_times_from_admission)
         radiology_note_times_from_admission = (all_radiology_texts, all_radiology_times_from_admission)
-
-        # icd_codes: (List[List[str]], List[float]) — codes per visit, hours from first admission
-        icd_codes = (all_icd_codes, all_icd_times)
 
         return [
             {


jhnwu3 (changes requested):
Ahh, some quick comments, because now I think I know what's happening that's happened to me before.

What you can probably do is just make sure every patient has a List[str]. For patients without a type of note, you can just append` "<missing>"` to denote a missing note or something of that sort or "". We'll probably have to standardize.

Another thing we can do is make sure to align our definitions with Rian's tuple time processors:

https://pyhealth.readthedocs.io/en/latest/api/processors/pyhealth.processors.TupleTimeTextProcessor.html

So you don't have to define a times feature.

i.e instead of

`input_schema = "discharge_times" : 'tensor', "discharge" : 'raw'`
what you can do is do:

`input_schema = "discharge_note_times" : "tuple_time_text"`

where each discharge_note_times is a (notes, times)

Let me know if this helps!

pyhealth/tasks/ehr_foundational_model_mimic4.py
```diff
@@ -0,0 +1,156 @@
...
+        if len(admissions_to_process) == 0:
+            return []
+
+        # Get first admission time as reference for lab time calculations
+        first_admission_time = admissions_to_process[0].timestamp
+
+        # Aggregated data across all admissions
+        all_discharge_notes = []  # List of individual discharge notes
+        all_radiology_notes = []  # List of individual radiology notes
+        all_discharge_notes_timestamps = [] # List of individual discharge notes timestamps
```
jhnwu3: It's also possible that our timestamps may not be being correctly computed? I'd have think about it.

pyhealth/tasks/ehr_foundational_model_mimic4.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self):
+        """Initialize the EHR Foundational Model task."""
+        self.input_schema: Dict[str, str] = {
+            "discharge": "raw",
+            "radiology": "raw",
+            "discharge_note_time_diffs": "tensor",
+            "radiology_note_time_diffs": "tensor",
+        }
+        self.output_schema: Dict[str, str] = {"mortality": "regression"}
+
```
jhnwu3, line 31: Found another bug, mortality should be a "binary" variable for our purposes here.

pyhealth/tasks/ehr_foundational_model_mimic4.py
```diff
@@ -0,0 +1,156 @@
...
+                try:
+                    note_text = self._clean_text(note.text)
+                    if note_text:
+                        all_radiology_notes.append(note_text)
+                        all_radiology_notes_timestamps.append(note.timestamp)
+                except AttributeError:
+                    pass
+
+        # Sort discharge_notes by timestamp 
+        all_discharge_notes_timestamps.sort()
```
jhnwu3: We should probably be careful here with our sort, since we want to sort the notes and timestamps in pair.

pyhealth/tasks/ehr_foundational_model_mimic4.py
will-pang: Ah yes, haha good catch! I made it into `regression` for testing because I was getting this error:

```
Traceback (most recent call last):
  File "/Users/wpang/Desktop/PyHealth/examples/foundation_ehr/multimodal_task.py", line 35, in <module>
    samples = dataset.set_task(task, cache_dir=f"{CACHE_DIR}/task", num_workers=8)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wpang/Desktop/PyHealth/pyhealth/datasets/base_dataset.py", line 936, in set_task
    builder.fit(dataset)
  File "/Users/wpang/Desktop/PyHealth/pyhealth/datasets/sample_dataset.py", line 167, in fit
    processor.fit(samples, key)
  File "/Users/wpang/Desktop/PyHealth/pyhealth/processors/label_processor.py", line 25, in fit
    raise ValueError(f"Expected 2 unique labels, got {len(all_labels)}")
ValueError: Expected 2 unique labels, got 1
```

I wonder if it was because in the `dev` sample I was working with haha all the patients had a mortality label of 0, so for testing purposes I just made it a regression task. I've made the update!

pyhealth/processors/tuple_time_text_processor.py
jhnwu3, line 1: We can just do to get pass the serialization problems:
`texts, time_diffs = value
        time_tensor = torch.tensor(time_diffs, dtype=torch.float32)
        return pickle.dumps(texts), time_tensor`

However, I think we may need to think of a better approach here, one that includes a tokenizer in our processor here.

jhnwu3 (changes requested):
Some other things we definitely need to revisit and follow now that I understand how the processors work, but fortunately PyHealth is really flexible so it should be doable:

1. In our Task, we'll need explicitly define the processor classes themselves with arguments https://pyhealth.readthedocs.io/en/latest/api/processors.html -> this documentation should explain how to define processor arguments in a task
3. The TupleTimeTextProcessor will need to leverage a HuggingFace Tokenizer so all texts will be tokenized into a [T x L] tensor of tokens, with a time tensor of Tdimension.
4. The TimeImageProcessor I think fortunately works as it should with the litdata expectations, just two tensors.
5. The TextEmbedding model will need to assume inputs are already tokenized

pyhealth/processors/tuple_time_text_processor.py
will-pang: Adding some notes from our convo in Discord so I can keep track of things:

Screenshot 2026-02-15 at 4 10 24 PM


Screenshot 2026-02-15 at 3 31 35 PM

will-pang:
I still have a few outstanding questions on computing `time_diff` for @jhnwu3 (I think we were debugging the processors so ran out of time to address this), but wanted to check on computing `time_diffs`:

1. Should I add a `.sort(key=lambda x: x[1])` on the (text, timestamp) lists before computing time diffs? Seems like MIMIC uses fake dates, but perhaps the fake dates still preserve chronology so I wanted to check with you haha

2. Anchor time vs. sequential diffs: Currently _compute_time_diffs computes `time_diff` from previous
  note. Should we instead use an anchor time (e.g., first admission time) so that diffs represent absolute offsets?
  
3. When a patient has zero notes of a given type, we return `(["<missing>"], [0.0])`. Is this correct?

**_Adding John's Reply on Discord:_**

Screenshot 2026-02-15 at 2 46 34 PM

jhnwu3 (changes requested):
- Rename the tasks/ehr_foundational_model_mimic4.py -> multimodal_mimic4.py
- Add docs/ 
- example script on how to run this

PR #841, opened 2026-02-11
Fix Chefer and add standardized API

Logiquo:
This PR furthur enhance the interpretability API

1. Move interpretability related API to a separate ABC `Interpretable` for model want to be interpretable to inherite, this make the BaseModel ABC cleaner
2. Create a new API called `CheferInterpretable` for Chefer specific API
3. Remove Chefer model specific attribution implementation, and use Chefere specific API to fetch information instead.

# Model
- [x] Transformer
- [x] StageAttn
- [ ] ~~ViT~~ (will be defered to later PR)

# Methods
- [x] Chefer

### DIFF
--- a/examples/interpretability/dka_stageattn_mimic4_interpret.py
+++ b/examples/interpretability/dka_stageattn_mimic4_interpret.py
@@ -134,7 +134,7 @@ def main():
         "ig_gim": IntegratedGradientGIM(model),
         "deeplift": DeepLift(model, use_embeddings=True),
         "gim": GIM(model),
-        # "chefer": CheferRelevance(model),
+        "chefer": CheferRelevance(model),
         "shap": ShapExplainer(model, use_embeddings=True),
         "lime": LimeExplainer(model, use_embeddings=True, n_samples=200),
     }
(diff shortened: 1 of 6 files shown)


pyhealth/interpret/methods/chefer.py
```diff
@@ -122,246 +129,328 @@ class CheferRelevance(BaseInterpreter):
...
+    return attn
+
+
+def _reference_attribute_vit(
+    model,
+    interpolate: bool = True,
+    class_index=None,
+    **data,
+) -> Dict[str, torch.Tensor]:
+    """[REFERENCE ONLY] Original ViT-specific Chefer attribution.
```
jhnwu3, line 396: Just a clarification question, I guess it's because our TorchVision wrapper doesn't/can't support a CheferInterpretable here due to technically not being a static model? 

https://github.com/sunlabuiuc/PyHealth/blob/master/pyhealth/models/torchvision_model.py

Logiquo: Due to i haven't had time to modify the ViT yet🤣, technically nothing stop it from happening. but given that we are working on interpretability project more, i want to fix the most relavent model first.

PR #869, opened 2026-02-23
feat: add streaming vs in-memory loader benchmark

AhmedAli58:


Logiquo:
Close with the same reason as https://github.com/sunlabuiuc/PyHealth/pull/871

PR #876, opened 2026-02-27
tldr; add improved check to task and dataset cache

jhnwu3:
This pull request introduces improvements to cache handling and data configuration in the `pyhealth` dataset codebase. The most important changes are grouped below by theme.

Cache handling improvements:

* Improved the cache validation logic in `global_event_df` by checking if the cache directory contains any `.parquet` files, and added logic to remove and rebuild the cache if the directory exists but is incomplete. This prevents issues with partial or corrupted cache states.

Data configuration updates:

* Added the `row_id` attribute to the `tables` section in the `mimic3.yaml` configuration, ensuring that this identifier is included in downstream processing.

### DIFF
--- a/tests/core/test_caching.py
+++ b/tests/core/test_caching.py
@@ -107,6 +107,9 @@ class TestCachingFunctionality(BaseTestCase):
         self.dataset = MockDataset(cache_dir=self.temp_dir.name)
         self.task = MockTask()
 
+    def tearDown(self):
+        self.temp_dir.cleanup()
+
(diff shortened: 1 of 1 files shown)


EricSchrock (changes requested):
Code looks good but I'd recommend adding a test case to `test_caching.py`. I'm guessing there will be further tweaks to caching and a good set of regression tests will be really helpful.

Happy to switch this to approve if you need to get this in quick or happy to add the regression test over the weekend if you're busy.

PR #892, opened 2026-03-14
Add SleepWakeClassification task for DREAMT

diegofariasc:
****Contributor:** Diego Farias Castro (diegof4@illinois.edu)
Type of contribution:** task
**Link to original paper:** https://proceedings.mlr.press/v248/wang24a.html

**High-level description:**
This PR adds a new standalone task, `SleepWakeClassification`, on top of the existing `DREAMTDataset`. The task supports epoch-level sleep-vs-wake prediction from multimodal wrist-worn wearable signals. It turns each DREAMT record into fixed-length epochs, extracts features from accelerometer, temperature, blood volume pulse, and electrodermal activity signals, adds temporal context features, and assigns a binary sleep/wake label to each epoch.

**Implementation summary:**
- Adds `SleepWakeClassification` in `pyhealth/tasks/sleep_wake_classification.py`
- Exports the task in `pyhealth/tasks/__init__.py`
- Adds task API documentation in `docs/api/tasks/pyhealth.tasks.sleep_wake_classification.rst`
- Registers the task page in `docs/api/tasks.rst`
- Includes an example ablation script in `examples/dreamt_sleep_wake_classification_lightgbm.py`
- Includes task tests in `tests/core/test_sleep_wake_classification.py`

**Reproducibility scope:**
This PR focuses on the task side of the paper. It makes the sleep-wake prediction setting available inside PyHealth so the generated samples can be used in new experiments and ablation studies.

**Task behavior:**
- Input: wearable records from the existing `DREAMTDataset`
- Output: epoch-level samples with `patient_id`, `record_id`, `epoch_index`,
  `features`, and binary `label`
- Labels: wake maps to `1`; sleep stages (`REM`, `N1`, `N2`, `N3`) map to `0`
- Features: accelerometer summaries, temperature summaries, BVP-based HRV
  features, and EDA-based SCR features
- Temporal context: Gaussian smoothing, temporal derivative, and rolling
  variance for each base feature

**File guide:**
- `pyhealth/tasks/sleep_wake_classification.py`: task implementation
- `pyhealth/tasks/__init__.py`: public task export
- `docs/api/tasks/pyhealth.tasks.sleep_wake_classification.rst`: task docs
- `docs/api/tasks.rst`: task index update
- `examples/dreamt_sleep_wake_classification_lightgbm.py`: example and ablation workflow
- `tests/core/test_sleep_wake_classification.py`: task unit tests


tests/core/test_sleep_wake_classification.py
```diff
@@ -0,0 +1,295 @@
+import numpy as np
+import pandas as pd
+
+from pyhealth.tasks.sleep_wake_classification import SleepWakeClassification
+
+
+class FakeEvent:
```
jhnwu3: I've been thinking a lot about standardizing where/what some test cases would typically look like here. I like how you reverse-engineering some aspects of what goes in within PyHealth, but I think it would be much better if we followed some of the other examples. (Still working on test case best practices here as we learn what seems to scale better and what doesn't).

But, if possible instead of constructing object oriented test classes, can we just create fake tmp data, and use the explicit datasets/data types in PyHealth?

See this example:
https://github.com/sunlabuiuc/PyHealth/blob/master/tests/core/test_chestxray14.py 

@EricSchrock can probably give way better advice on this. But, the tldr; we want the testing environment to mimic the real working environment as closely as possible here.

pyhealth/tasks/sleep_wake_classification.py
```diff
@@ -0,0 +1,886 @@
...
+
+from ..data import Patient
+from .base_task import BaseTask
+
+
+class SleepWakeClassification(BaseTask):
+    """Binary sleep-wake classification task for DREAMT wearable recordings.
+
+    This task converts each DREAMT wearable recording into fixed-length epochs,
+    extracts physiological features from multiple sensor modalities,
```
jhnwu3, line 56: Can you give a rough breakdown of what's happening here? It's a bit difficult to follow this like 10 different private functions haha. It'd be good to have it flow, something like starting from step 1 to step 9, this is what's going on, and the sequence of function calls here. 

Similarly, in the doc strings, can you add an example on how a user might use DreamT here with the set_task() call here.

jhnwu3: It's also not clear exactly what we mean by features here. What shape is it? What do its dimensions mean?


I think a key thing about what we want to do as good practice is to explain what it is we're doing here and what are the expected outputs here.

examples/dreamt_sleep_wake_classification_lightgbm.py
```diff
@@ -0,0 +1,376 @@
...
+    X_test = X_test[:, non_all_nan_cols]
+
+    print(f"{BOLD}Feature count:{RESET} {X_train.shape[1]}")
+
+    imputer = SimpleImputer(strategy="median")
+    X_train = imputer.fit_transform(X_train)
+    X_test = imputer.transform(X_test)
+
+    # Train a LightGBM model on the current feature subset.
+    train_data = lgb.Dataset(X_train, label=y_train)
```
jhnwu3, line 194: Just curious, is it possible to create a ContraWR example with a signals-based model here? The lightgbm isn't bad, and I see the inputs in essence are functionally formatted like a table in some sense (table of signals) here, which is still pretty cool to see as an example. 

But this is a signals dataset after all:
https://physionet.org/content/dreamt/2.1.0/

jhnwu3 (changes requested):
Let me know if you have any more questions. Nice work!

tests/core/test_sleep_wake_classification.py
diegofariasc: Thanks, this makes sense. I updated the tests to use a small temp DREAMT-like CSV plus real PyHealth Patient/Event plumbing instead of custom FakePatient/FakeEvent classes. The expensive feature extraction is still mocked where needed, so the tests stay fast

pyhealth/tasks/sleep_wake_classification.py
diegofariasc: Thanks for the comments. I just added a clearer task docstring with the step-by-step flow and a short DREAMTDataset.set_task(SleepWakeClassification()) example.

diegofariasc: Also, I added more clarification in the task docstring. Here, features means the numeric summary we compute for one sleep epoch, not the raw wearable signal. For each 30-second epoch, the task creates one flat list with 84 numbers. The first 21 numbers summarize the ACC, TEMP, BVP, and EDA signals. The remaining 63 numbers add temporal context: for each of those 21 base values, I add a smoothed value, a derivative, and a rolling variance. So the task returns one sample per labeled epoch with patient_id, record_id, epoch_index, features, and binary label.

examples/dreamt_sleep_wake_classification_lightgbm.py
diegofariasc: I think so. ContraWR would make sense here because DREAMT is a signal dataset. The current example is using the task’s extracted epoch features, so the data going into the model is table-like rather than raw waveform. A ContraWR example would probably need a raw-signal epoch pipeline instead of the 84-feature vector this task emits. I did not add that here, but maybe I can open a new follow-up PR with those changes.

PR #894, opened 2026-03-15
[Conformal EEG] TUEV/TUAB Compatibility

lehendo:
Arjun Chatterjee

Implemented the fixes John and I discussed to make PyHealth compatible with TUEV and TUAB. This should work with ContraWR and TFM-Tokenizer now. I also added and updated the example scripts to run the conformal prediction methods on this updated repository.


pyhealth/tasks/temple_university_EEG_tasks.py
```diff
@@ -139,40 +139,60 @@ def readEDF(fileName: str,
     def __call__(self, patient: Any) -> List[Dict[str, Any]]:
```
jhnwu3, line 144: Can we create a boolean for stft=True here? So it's not always computed in case people just want the raw signal too.

PR #897, opened 2026-03-17
[SyntheticEHR] HALO Baseline

jalengg:
See notes for #878

Notebook permalink: https://colab.research.google.com/github/jalengg/PyHealth/blob/halo-pr-integration/examples/halo_mimic3_colab.ipynb


examples/halo_mimic3_training.py
```diff
@@ -0,0 +1,66 @@
...
+)
+base_dataset.stats()
+
+# Step 2: Set task for HALO generation
+# halo_generation_mimic3_fn extracts diagnosis code sequences per patient.
+# Each patient produces one sample with all their visits (admissions with
+# at least one ICD-9 code). Patients with fewer than 2 qualifying visits
+# are excluded.
+print("Setting HALO generation task...")
+sample_dataset = base_dataset.set_task(halo_generation_mimic3_fn)
```
jhnwu3, line 33: Can we use the TaskClass you wrote here for this instead of the function that I can't see anymore?

pyhealth/models/generators/halo.py
```diff
@@ -0,0 +1,378 @@
...
+                    batch_size=self._batch_size,
+                    shuffle=False,
+                    drop_last=False,
+                    collate_fn=_halo_collate_fn,
+                )
+                val_losses = []
+                with torch.no_grad():
+                    for val_batch in val_loader:
+                        visits = val_batch["visits"].to(self.device)
+                        batch_ehr, batch_mask = self._encode_visits(visits)
```
jhnwu3, line 262: I see there's an encode function. This can be quite prohibitively expensive in training if it's some form of tokenization here or whatever.

I guess the question is that is batch_ehr here an embedding here? or is it a tokenization approach here? A collate function?

pyhealth/models/generators/halo.py
```diff
@@ -0,0 +1,378 @@
...
+
+    # ------------------------------------------------------------------
+    # forward — required by BaseModel (abstract in nn.Module)
+    # ------------------------------------------------------------------
+
+    def forward(self, visits: torch.Tensor, **kwargs) -> Dict[str, torch.Tensor]:
+        """Forward pass.
+
+        Accepts the padded index tensor produced by NestedSequenceProcessor,
+        converts it to HALO multi-hot format, and runs the transformer.
```
jhnwu3, line 168: If you need a multihot format, we have a multihot processor here. We should probably chat on some next steps.

jhnwu3 (changes requested):
I've made some small comments here, but will summarize more on Discord since I think you guys are actually really close. There's some quick discussions on what I think is arguably more efficient and effective. And honestly will confuse the users who build upon this work later about this.

PR #926, opened 2026-03-31
Fix Interpretability Methods `target_class_idx`

Logiquo:
1. Unify all interpretability to support `target_class_idx` and its semantics
2. `target_class_idx` becomes a no-op for binary classification, and select target class we want to interpret for multi-class and multi-label.
3. The multi-label will be selected based on `argmax` if `target_class_idx` is not provided. 
4. All samples (including negatives) will be evaluated in the interpretability metric, can support `sample_filter` which can specify POSITIVE, NEGATIVE and IGNORE based on the predicated probablity.

### DIFF
--- a/pyhealth/metrics/interpretability/evaluator.py
+++ b/pyhealth/metrics/interpretability/evaluator.py
@@ -35,9 +35,10 @@ class Evaluator:
         sample_filter: A callable that classifies each sample for evaluation.
             Signature: (class_probs, classifier_type) -> sample_classes
             where class_probs has shape (batch_size,) and contains the
-            probability for the predicted class (sigmoid/softmax output
-            with target class already applied), and sample_classes is a
-            tensor of SampleClass values:
+            class probability used for filtering. For binary single-logit
+            models, this is ``P(class=1)``. For multiclass/multilabel
+            models, this is the gathered target-class probability.
+            ``sample_classes`` is a tensor of SampleClass values:
             - SampleClass.POSITIVE: evaluate with attributions as-is
             - SampleClass.NEGATIVE: evaluate with negated attributions
             - SampleClass.IGNORE: exclude from evaluation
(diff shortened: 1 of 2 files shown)


pyhealth/metrics/interpretability/evaluator.py
```diff
@@ -30,17 +32,45 @@ class Evaluator:
...
+        ...     threshold_sample_filter,
+        ... )
         >>>
-        >>> # Initialize evaluator
+        >>> # Initialize evaluator with default filter
         >>> evaluator = Evaluator(model)
         >>>
+        >>> # Initialize with custom filter that ignores low-confidence
+        >>> def confident_filter(class_probs, classifier_type):
+        ...     batch_size = class_probs.shape[0]
```
jhnwu3, line 66: Just curious, how does this work? That's pretty cool though. Can we add examples here in the examples/interpretability/ directory that show how a user might do this for our binary cases here iirc?

pyhealth/interpret/methods/base_interpreter.py
```diff
@@ -207,6 +211,42 @@ def attribute(
...
+        logit to explain.  All prediction modes share this single code
+        path:
+
+        * **Binary** (single logit): ``target_class_idx`` is a no-op
+          because there is only one output.  Always returns zeros
+          (index 0).
+        * **Multi-class / multi-label**: uses ``target_class_idx`` if
+          given, otherwise the argmax of logits.
+
+        Args:
```
jhnwu3, line 231: I like this approach.

Logiquo:
image

image

We can have custom function to determine what sample is considered as positive/negative/ignore.

pyhealth/metrics/interpretability/evaluator.py
Logiquo: Added

PR #932, opened 2026-04-02
Add download source and required files documentation to datasets

### LINKED ISSUE #883: Add documentation to each dataset on download sources, and what files specifically are compatible.
Right now, we just assume the users know where each dataset is located, and exactly what files are required.

But, MIMIC-CXR vs. MIMIC-CXR-jpg are 2 different datasets referencing the same dataset. 

MIMIC-CXR-JPG is supported, but MIMIC-CXR is not.

sevvaluluss:
Hi! I'm a new contributor and wanted to help with issue #883.

I added download links and required file lists to the docstrings of these datasets:
- eICUDataset
- MIMIC3Dataset
- MIMIC4EHRDataset
- EHRShotDataset
- CardiologyDataset

This way users don't have to guess where to find the data or which files they need. Let me know if anything needs to be changed!

Closes #883

### DIFF
--- a/pyhealth/datasets/cardiology.py
+++ b/pyhealth/datasets/cardiology.py
@@ -10,7 +10,20 @@ from pyhealth.datasets import BaseSignalDataset
 
 class CardiologyDataset(BaseSignalDataset):
     """Base ECG dataset for Cardiology
+    
+    Download: https://physionet.org/content/challenge-2020/1.0.2/
+    Access: Requires PhysioNet account.
 
+    Required files:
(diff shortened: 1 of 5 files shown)


jhnwu3 (approved):
Hey, thanks for the help! 

It doesn't entirely fix everything, but it's a really good start!

We're also thinking of adding a download=False option sometime in the future. There's a couple challenges with that, mainly many of the dataset download services can be really really slow, and testing this can be a little challenging here, so if you're interested on working on this feature, would be happy to chat further on the discord.

PR #940, opened 2026-04-04
Daily and Sports Activities Dataset & Task

ranyou:
I've implemented the dataset & task for Daily and Sports Activities (DSA) dataset. 

The Daily and Sports Activities dataset contains motion sensor data. 

More details here: https://archive.ics.uci.edu/dataset/256/daily+and+sports+activities

This dataset was used by paper "Daily Physical Activity Monitoring: Adaptive Learning from Multi-source Motion Sensor Data" https://proceedings.mlr.press/v248/zhang24a.html


pyhealth/tasks/activity_classification.py
```diff
@@ -0,0 +1,98 @@
...
+        >>> from pyhealth.tasks import ActivityClassification
+        >>> dataset = DSADataset(root="/path/to/dsa")
+        >>> task = ActivityClassification()
+        >>> samples = dataset.set_task(task)
+    """
+
+    task_name: str = "ActivityClassification"
+    input_schema: Dict[str, str] = {
+        f"{x}_{y}{z}": "sequence"
+        for x in ["T", "LA", "RA", "LL", "RL"]
```
jhnwu3, line 38: Are there really that many sequential features here? That's a pretty wide input space.

jhnwu3 (commented):
Leaving a comment here as it looks pretty good.

pyhealth/tasks/activity_classification.py
jhnwu3: Also, a sequence in this case is a "sequence" of discrete variables, you might want to use a tensor processor here and just directly convert to a tensor vector if there's a lot of raw continuous values through time here.

jhnwu3 (changes requested):

PR #946, opened 2026-04-05
[CPBench] Benchmarking Wrapper (Old, Not Final)

lehendo:
Adds pyhealth/calib/benchmark.py with a CalibrationBenchmark class that evaluates a conformal prediction algorithm against all built-in baselines in one call.


jhnwu3 (commented):
Poking @siddharthal hopefully he sees this. Tbh, It's not really clear how a benchmark wrapper should look like here. 

We also probably need to think very critically or draft up at some point a good way of organizing the docs to automatically mirror and be easy to navigate in the future as the number of tasks and datasets supported increases here.

PR #957, opened 2026-04-09
Add SleepStagingDREAMT task for DREAMT wearable sleep dataset

ErieMAdames:
**Authors**: Jash Parekh (jashrp2@illinois.edu), Tom Wang (twwang2@illinois.edu), Erie M. Adames (eadames2@illinois.edu)
**Contribution Type**: Standalone Task (Option 3)
**Paper**: Wang et al., "DREAMT: Dataset for Real-time Sleep Stage Estimation using Multisensor Wearable Technology", CHIL 2024
**Paper Link**: https://proceedings.mlr.press/v248/wang24a.html

## Overview

This PR adds a sleep staging task (`SleepStagingDREAMT`) for the existing DREAMT dataset in PyHealth. The task segments overnight Empatica E4 wearable signals into 30-second epochs and assigns sleep stage labels with support for 2-class (wake/sleep), 3-class (wake/NREM/REM), and 5-class (W/N1/N2/N3/R) granularity.

## Task (`sleep_staging_dreamt.py`)

- Processes 8 E4 signal channels (ACC x/y/z, BVP, EDA, HR, IBI, TEMP)
- Applies signal-specific preprocessing (Butterworth bandpass for ACC, Chebyshev highpass for BVP/EDA, winsorization for TEMP)
- Supports configurable signal column subsets for ablation studies
- Maps PSG-annotated sleep stages to 2, 3, or 5 classes
- Excludes artifact epochs (stage "P") and missing data
- Properly inherits from `BaseTask` with `input_schema` and `output_schema`

## Ablation Study (`examples/dreamt_sleep_wake_lstm.py`)

Two ablation studies with 5-fold participant-level cross-validation:

1. **Signal subset ablation** — compares ACC-only, BVP+HRV, EDA+TEMP, and ALL signals using 2-class labels
2. **Label granularity ablation** — compares 2-class (wake vs sleep) vs 5-class (W/N1/N2/N3/R) using all signals

- Includes a single-layer LSTM classifier
- Runnable with `--demo` flag using fully synthetic data (no dataset download required)
- Reports F1, AUROC, Accuracy, and Cohen's Kappa

## Unit Tests (`tests/core/test_sleep_staging_dreamt.py`)

- 26 tests using synthetic CSV data via pytest `tmp_path`
- Covers: initialization, label mappings (2/3/5-class), signal shape, patient ID, epoch ordering, stage exclusion (P / Missing), signal column subsets, filtering/winsorization, and edge cases
- All tests complete in ~3-5 seconds using only synthetic data

## File Guide

| File | Description |
|------|-------------|
| `pyhealth/tasks/sleep_staging_dreamt.py` | Core task implementation |
| `pyhealth/tasks/__init__.py` | Added `SleepStagingDREAMT` import (1 line) |
| `tests/core/test_sleep_staging_dreamt.py` | 26 unit tests (synthetic data only) |
| `examples/dreamt_sleep_wake_lstm.py` | LSTM ablation study script with `--demo` mode |
| `docs/api/tasks/pyhealth.tasks.SleepStagingDREAMT.rst` | API documentation stub |
| `docs/api/tasks.rst` | Added entry to Available Tasks toctree (1 line) |


joshuasteier:
Hi Jash, Tom, and Erie(@ErieMAdames). The task correctly inherits from `BaseTask`, the signal-specific preprocessing (Butterworth for ACC, Chebyshev for BVP, winsorization for TEMP) follows the paper's methodology, the test coverage with 26 synthetic tests is thorough, and the example script includes a working `--demo` mode that runs without real data. Patient-level splitting via `split_by_patient` is correct for sleep staging. A few things worth addressing.

**Label assignment after row-filtering excluded stages.**

In `__call__`, you first drop all rows with excluded stages:

```python
mask = ~df[stage_col].isin(EXCLUDED_STAGES)
df = df.loc[mask].reset_index(drop=True)
```

Then you epoch the filtered DataFrame in 30-second blocks and take the label from the middle row:

```python
mid = start + self.epoch_len // 2
stage = str(df[stage_col].iloc[mid]).strip()
```

If excluded rows were removed from the middle of what was originally one labeled PSG epoch, the remaining rows collapse together. A contiguous 1920-sample block in the filtered frame can span what were originally two different stages, and it still gets a single label from the middle index. The signal for that epoch would be spliced.

A more standard approach is to epoch first (based on original timestamps or row index), then for each epoch check whether its majority-label is excluded and drop the whole epoch if so. This preserves the 30-second signal coherence within an epoch. Could you rework this, or add a test that constructs an input where row-filtering straddles an epoch boundary and verifies the behavior you want?

**The example throws away the temporal structure.**

Your task outputs rich `(n_channels, 1920)` tensors at 64 Hz. The example then reduces each epoch to 4 summary statistics per channel (mean, std, min, max) and feeds a flat `4 * n_channels` vector to PyHealth's RNN:

```python
def _epoch_features(signal: np.ndarray) -> List[float]:
    for ch in range(signal.shape[0]):
        s = signal[ch].astype(np.float64)
        feats.extend([float(np.mean(s)), float(np.std(s)),
                      float(np.min(s)), float(np.max(s))])
    return feats
```

This defeats most of the point of having 30-s raw signal windows, and it means a user following the example is not actually running a sleep-staging model on the raw signal. PyHealth has `CNN` and `SparcNet` that are better suited for multichannel 64 Hz windows, and they would make this a realistic baseline. Could you either switch the example to use a model that consumes the raw signal, or add a second example that does?

If you keep the feature-extraction version, please explain in the docstring why you chose summary stats rather than raw signals.

**PR description mentions 5-fold CV but the script does 70/10/20.**

The PR description says "Two ablation studies with 5-fold participant-level cross-validation." The script does one 70/10/20 patient-level split. Could you either update the description to match the script, or run actual 5-fold CV? The script approach is fine as long as the description is accurate.

**Smaller items**

1. `get_events` fallback:
   ```python
   try:
       events = patient.get_events(event_type="dreamt_sleep")
   except (TypeError, KeyError):
       events = patient.get_events()
   ```
   The fallback suggests uncertainty about `DREAMTDataset`'s event type. Could you verify the actual event type and remove the fallback, or add a comment explaining why both paths are needed?

2. `super().__init__()` is called at the end of `__init__` with no arguments. Some `BaseTask` implementations expect attributes to be set before `super().__init__()` runs initialization logic, and others expect the opposite. Could you verify this is the intended order?

3. `TIMESTAMP` column is present in test CSVs but never validated. If the file's actual sampling rate differs from `sampling_rate=64`, the filters will be wrong. Consider a sanity check that compares expected row count to `(last_timestamp - first_timestamp) * sampling_rate` and logs a warning on mismatch.

4. Test `test_p_stage_excluded` asserts `all(lbl in {0, 1, 2, 3, 4} for lbl in labels)`, which is true by construction since the label map only produces those values. A stronger test would assert the expected label count, for example `len(samples) == 2` for input `["P", "P", "W", "N1"]`.

5. The example's `run_config` creates the dataset, splits, builds the model, trains, and evaluates in one function. Splitting this into `build_dataset`, `train`, and `evaluate` would make the flow easier to follow and easier to reuse in other scripts.

PR #979, opened 2026-04-14
Add RetinaUNet model into Pyhealth, expand usage with sample notebook.

iamkishann:
## Contributors
- Kishan Sarvaiya (NetID: kishans3) ([kishans3@illinois.edu](mailto:kishans3@illinois.edu))
- Victor Josue Rosales Jimenez (NetID: victorr5) ([victorr5@illinois.edu](mailto:victorr5@illinois.edu))

## Type
Model Contribution (Option 2)

## Paper
Retina U-Net: Embarrassingly Simple Exploitation of Segmentation Supervision for Medical Object Detection — [arXiv:1811.08661](https://arxiv.org/abs/1811.08661)

---

## Description

This PR implements `RetinaUNet`, a multi-task model for medical image object detection and segmentation. The model combines a Feature Pyramid Network (FPN) backbone with Retina-style anchor-based detection heads and a U-Net-style segmentation head. It is wrapped as a PyHealth `BaseModel` and supports both 2D and 3D inputs.

## Key Features

- **Dual-task learning**: joint object detection and segmentation in a single forward pass
- **2D and 3D support**: configurable via the `dim` parameter
- **Anchor-based detection**: multi-scale anchors over FPN levels P2–P5 with sub-scaling
- **Built-in post-processing**: bounding box delta decoding, clipping, and class-wise NMS
- **PyHealth integration**: inherits from `BaseModel`; returns standard loss and prediction keys (`loss`, `class_loss`, `bbox_loss`, `seg_loss`, `det_bboxes`, `class_logits`, `bbox_deltas`, `seg_logits`, `anchors`)

## Files to Review

1. `pyhealth/models/retina_unet.py` — Main implementation
2. `tests/core/test_retina_unet.py` — Test suite
3. `docs/api/models/pyhealth.models.RetinaUNet.rst` — Documentation
(add notebook) - ablation study

## Testing

4 unit tests pass using synthetic data:

| Test | Description |
|---|---|
| `test_model_initialization` | Verifies submodules and config |
| `test_model_forward` | Verifies training forward pass keys, shapes and scalar loss |
| `test_model_accepts_missing_labels_in_eval_mode` | Verifies inference without labels yields zero losses and correct output shapes |
| `test_model_backward` | Verifies gradient computation flow through all parameters |

Please squash commits when merging - select the squash all commits option


joshuasteier:
@iamkishann:
Hi Kishan and Victor, please see feedback below:

**Bbox delta encoding does not match delta decoding.**

In `_compute_anchor_matches`:

```python
anchor_target_deltas = (matched_gt_boxes - anchors) / (anchors + 1e-5)
```

This is a naive elementwise ratio on raw box coordinates. In `_apply_box_deltas_2d` (used at inference), you decode using the standard Faster-RCNN formulation: centre offsets normalised by anchor size, and log-ratios for height/width. These two operations are not inverses, so the regression targets you train on are not the targets the decoder expects. The box head will not converge correctly on real data.

The correct 2D encoding is:

```python
dy = (gt_cy - anchor_cy) / anchor_h
dx = (gt_cx - anchor_cx) / anchor_w
dh = log(gt_h / anchor_h)
dw = log(gt_w / anchor_w)
```

For 3D add `dz` and `dd` analogously. There is also a second scale mismatch: at inference, predicted deltas are multiplied by `self.rpn_bbox_std_dev` (`[0.1, 0.1, 0.2, 0.2]`), but the training targets are not divided by those same values, so the two sides are on different scales even if the encoding were otherwise correct.

**NMS threshold is `1e-5`.**

```python
self.nms_threshold = 1e-5
```

This is essentially zero IoU tolerance. In `_nms_2d` the check is `iou <= iou_threshold`, so any two boxes with any overlap at all suppress one. Standard NMS thresholds are in the range 0.3 to 0.7. Please set a sensible default (0.5 is typical for detection).

**Anchor matching is hardcoded to 2D even when the core is 3D.**

`_compute_anchor_matches` takes a `dim` argument with default 2, but the call site never passes it:

```python
anchor_class_match, anchor_target_deltas = self._compute_anchor_matches(
    anchors, gt_boxes, gt_class_ids
)
```

So when the model is instantiated with `dim=3`, the loss path falls through to `_compute_iou_matrix_2d` on 6-column boxes, which will compute an incorrect IoU (it treats `z1`, `z2` columns as if they were part of the 2D box). Please propagate `self.core.dim` into the call.

**Class naming inconsistency.**

The commit "Refactor: Rename RetinaUNetCore to RetinaUNetLayer" landed, but the actual class is still `RetinaUNetCore` in the code, and `__init__.py` imports `RetinaUNetCore`. Your PR description and the `RetinaUNet` docstring also mention `RetinaUNetLayer`. Please either complete the rename to `RetinaUNetLayer` (matches repo convention of `RETAINLayer`, `ConCareLayer`, etc.) or revert the description and docstrings to `RetinaUNetCore`.

**Inference mode silently returns `loss=0`.**

When `gt_seg_masks` is None, all three loss components are set to `torch.tensor(0.0)` and the returned `"loss"` is a zero tensor with no gradient. If a user forgets to pass labels during training, the model will train toward zero loss and appear "fine." Please add an explicit check: if the model is in training mode (`self.training == True`) and labels are missing, raise a `ValueError` instead of silently returning zero loss.

**`test_model_inference_without_labels` is misleading.**

The test passes only `images` then asserts `loss`, `class_loss`, etc. are present. Those keys are always in the output (with zero values), so the test is trivially satisfied. Either rename to something like `test_model_accepts_missing_labels` or change the assertions to verify actual inference behaviour (for example, that `det_bboxes` has the right shape and that `loss` is zero or absent once you change the behaviour per the point above).

**The `.gitignore` changes are out of scope.**

Your diff adds:

```
.vscode/
examples/dummy_data*/
examples/checkpoints*/
examples/logs*/
examples/__pycache__/
examples/**/__pycache__/
weightfiles/
```

These are personal development artifacts. The repo's `.gitignore` shouldn't change to accommodate this PR's workflow. Please revert the `.gitignore` changes and handle these paths in your local gitignore or a global `~/.gitignore`.

**Hardcoded detection constants should be configurable.**

```python
self.pre_nms_limit = 3000
self.nms_threshold = 1e-5
self.max_instances_per_batch_element = 10
```

These are set inside `RetinaUNetCore.__init__` with no way to override. Please expose them as constructor arguments (and set `nms_threshold=0.5` or similar as the default per the earlier point).

**`torch.cuda.empty_cache()` inside anchor cache management.**

`_clear_anchor_cache` and the LRU eviction path both call `torch.cuda.empty_cache()` unconditionally. Libraries generally shouldn't manage CUDA's caching allocator on behalf of users because it has performance consequences (forces reallocation and slows subsequent forward passes). Please remove these calls and let users manage memory explicitly if they need to.

**Segmentation head hardcoded to binary.**

`RetinaUNetCore.__init__` sets `self.num_classes_seg = 2` with no way to configure it. This is fine for the paper's setup but should be documented in the `RetinaUNet` docstring so users understand that the `num_classes` argument only controls detection classes, not segmentation.

**Smaller items**

1. 55 commits is messy. Before merge, please squash to a small number of logical commits (for example, "Add RetinaUNet model and core," "Add tests," "Add docs").

2. The `if __name__ == '__main__':` block at the end of `retina_unet.py` prints model output shapes on direct execution. This is not necessary in a library module. Consider moving it to a script in `examples/` or removing it.

3. The test `test_model_backward` only asserts that at least one parameter has a gradient. Given the size of the model (FPN plus three heads), it would be stronger to assert that representative submodules (the FPN stem, the classification head, the seg head) all receive gradients.

4. The notebook is 1,397 lines and not rendered in the diff. If you could add a short "Results" markdown section at the top with the key numbers from the ablation, reviewers can see the summary without running the whole notebook.

5. The model bypasses PyHealth's `get_loss_function`, `prepare_y_prob`, and `get_output_size`. Object detection is genuinely not a standard classification task, so this is defensible, but a comment in the `RetinaUNet` class docstring explaining why these `BaseModel` helpers are not used would help future readers.

iamkishann:
@joshuasteier thanks for your reply, we have a few questions 

regarding the summary of notebook in a markdown, did you mean a cell on top of notebook as a markdown ? because we added a markdown cell on top of notebook and it seems like it does not render on github. the other option would be to add a separate summary markdown file for that but then that would not follow the schema

PR #986, opened 2026-04-14
Add GRU-D model for mortality prediction.

tracy030115:
**Contributor:** Tracy Zhang (tracyz3@illinois.edu)
**Contribution Type:** Model (Reproducibility and ablation)
**Paper:** https://arxiv.org/abs/1908.00690
**Description:** Implemented GRU-D (Gated Recurrent Unit with Decay), a RNN designed for irregularly sampled clinical time series with missing values. GRU-D extends the standard GRU with two decay mechanisms: input decay and hidden state decay. The model compared Raw, CUI, and Clinical feature representations along with hyperparameter tuning and ablation on the two decay mechanisms. 
**Files to Review:**
 - `pyhealth/models/grud.py`
- `pyhealth/models/__init__.py`
- `tests/core/test_grud.py`
- `docs/api/models/pyhealth.models.GRUD.rst`
- `docs/api/models.rst`
- `examples/mortality_prediction/mimic3_mortality_grud.ipynb`


joshuasteier:
Hi @tracy030115. The GRU-D math is faithful to Che et al. 2018, the ablation switches (`use_input_decay` and `use_hidden_decay`) are cleanly wired through, the test suite is unusually thorough, and the integration with `BaseModel` is correct. The notebook structure with three ablations is exactly the right scope. A few things worth addressing before merge.

**The `x_mean` API allows leakage even though your notebook avoids it.**

In your `train_one_fold` you correctly pass a train-only dataset to `GRUD(...)`, so `x_mean` is computed from training data only. The docstring even says so explicitly. However, the model API itself does not enforce this. A user doing the natural thing (`GRUD(dataset=full_dataset)` before splitting) would compute `x_mean` over train, val, and test. The paper requires this to come from training only.

Could you add an optional `x_mean` argument to the constructor so the good practice you already follow is discoverable from the API?

```python
def __init__(self, dataset, hidden_size=64, ..., x_mean=None):
    ...
    if x_mean is None:
        x_mean = self._compute_x_mean(dataset, first_key)
    # use x_mean
```

A warning in the docstring explaining that `x_mean` should be computed from the training split only (or computed on a dataset that contains only training samples) would help users avoid the footgun.

**Reproducibility on real data is unverified.**

The PR description describes this as a reproducibility and ablation contribution. The notebook runs with `USE_REAL_DATA = False` by default, all reported numbers (mortality AUROC ~0.56, long LOS ~0.48) are from 50-sample synthetic data, and the reported decay ablation shows hidden decay slightly hurting performance on synthetic. Your note correctly explains why this happens on synthetic data. The concern is that the PR never demonstrates the implementation actually reproduces Che et al. 2018 or Nestor et al. 2019 on real MIMIC-III.

This is not a blocker if the contribution is framed as "implementation + synthetic validation of the decay-switch mechanics," but the current framing is "reproducibility." Either:

1. Reframe the PR description and the notebook intro to say "implementation and ablation on synthetic data; real-data reproduction is left to users who have MIMIC-III access," or
2. Run the notebook on real MIMIC-III once and paste at least the decay ablation numbers from real data into the notebook (or the PR description) so reviewers can see `Full GRU-D > Standard GRU` in the setting where GRU-D is actually designed to shine.

**The `processing.py` preprocessing pipeline is referenced but not shipped.**

The notebook, the model docstring, and the test docstrings all refer to "the simple imputer pipeline in processing.py." No such file exists in this PR. Users who want to run GRU-D on real MIMIC-III data through PyHealth have no path to do so without reimplementing the preprocessing themselves. This is a significant usability gap for the "wide accessibility" goal of PyHealth.

Could you either (a) include `processing.py` in this PR, (b) add a helper function to `GRUD` (for example `GRUD.prepare_input`) that takes standard PyHealth time-series data and produces the interleaved format, or (c) at minimum add explicit step-by-step preprocessing instructions to the notebook so a user can start from MIMIC-III CSVs and reach the model?

**`GRUDLayer` contract vs how `GRUD.forward` uses it.**

In `GRUD.forward`:

```python
mask, mean_vals, delta = self._split_channels(x)
x_last_obsv = mean_vals.clone()
h = self.grud_layers[key](x=mean_vals, x_last_obsv=x_last_obsv, ...)
```

Both `x` and `x_last_obsv` are the same tensor. The `GRUDLayer` docstring says `x` contains "Raw measurements at the current timestep, shape ..., Contains NaN where missing" and `x_last_obsv` contains "Forward-filled measurements." In the model code they are the same thing (the forward-filled mean channel).

This is probably correct given the interleaved format assumes forward-fill has already happened upstream, but the layer-level docstring is misleading. Could you update the `GRUDLayer` docstring to describe the actual semantics (both inputs are forward-filled, with `mask` distinguishing observed from imputed positions)?

**`filter_square_matrix` as `Parameter` instead of buffer.**

```python
self.filter_square_matrix = nn.Parameter(
    filter_square_matrix, requires_grad=False
)
```

Non-learnable tensors that move with `.to(device)` should be registered via `register_buffer` rather than `nn.Parameter(requires_grad=False)`. The current form shows up in `model.parameters()` even though it has no gradients, which can confuse parameter counting and some optimizers. Same behavior, cleaner API.

**`forward` does not return `logit`.**

Other PyHealth models (`LogisticRegression`, `TransformerDeID`) return `{loss, y_prob, y_true, logit}`. `GRUD.forward` omits `logit`. Not strictly required by `BaseModel` but it is the convention. Adding `"logit": logits` to the returned dict is a one-line fix.

**Smaller items**

1. Typo in PR title: "predition" should be "prediction."

2. `from pyhealth.models.grud import GRUD` in `__init__.py` uses the absolute import path, while other models use relative imports like `from .grud import GRUD`. Trivial, but consistency.

3. `test_model_save_and_load` creates `model2 = make_model(make_synthetic_dataset(), hidden_size=8)` which generates a fresh dataset and therefore a fresh `x_mean`. The test passes because `x_mean` is deterministic under the seed, but reusing `self.dataset` for `model2` would make the test's intent cleaner.

Overall the implementation is solid and the ablation infrastructure is well built. The two things that matter most are making the `x_mean` API leak-resistant and either shipping the preprocessing pipeline or clearly documenting that real-data reproduction is left to the user.

tracy030115:
Thank you for the thorough feedback! I really appreciate it.

Changes:
- Added the decay ablation outputs from real data into the notebook as a table
- Updated `GRUDLayer` docstring to clarify that both `x` and `x_last_obsv` are forward-filled values, with `mask` distinguishing directly observed from imputed positions
- Changed `filter_square_matrix` from `nn.Parameter(requires_grad=False)` to `register_buffer()` so it moves with `.to(device)` for cleaner API
- Added `"logit": logits` to the returned dict
- Added optional `x_mean` argument to `GRUD.__init__` with a docstring warning explaining that it must be computed from training data
- Added `GRUD.prepare_input` static method that converts separate `values`, `mask`, and `delta` tensors into the interleaved format expected by the model
- Fixed `test_model_save_and_load` to reuse `self.dataset` for `model2` instead of generating a fresh dataset
- Changed from absolute import to relative import in `__init__.py`
- Fixed typo "predition" to "prediction" in PR title

joshuasteier:
Thanks @tracy030115.

Two things found while reading the file that I'd like you to look at before merge.

**`x_mean` time-dimension handling looks inconsistent.**

`_compute_x_mean` returns shape `(1, seq_len, input_size)` — preserves the time dimension by averaging only across samples:
```python
return stacked.mean(dim=0, keepdim=True)  # (1, seq_len, input_size)
```

But in `_decay_step` you index only the first timestep:
```python
x_mean_t = self.x_mean.squeeze(0)[0]  # (input_size,)
```

So the time-varying mean you compute is discarded — every timestep uses the mean from t=0 of the training data. The Che et al. paper treats `x̄` as a per-feature global mean (averaged across both samples and timesteps), so the simplest fix is to average over time as well in `_compute_x_mean`:
```python
return stacked.mean(dim=(0, 1), keepdim=True)  # (1, 1, input_size)
```
and update the buffer shape. Alternatively, if you want a time-varying mean (which is a deviation from the paper but plausible for clinical data), `_decay_step` needs to receive `t` and index `self.x_mean[0, t, :]`. Right now it does neither. The real-data results are still meaningful because the t=0 mean is a reasonable approximation, but the data flow doesn't match either intent cleanly.

**Docstring example references a method that doesn't exist.**

In the `x_mean` argument docstring:
```python
train_mean = GRUD._compute_x_mean_from(train_dataset, "time_series")
model = GRUD(dataset=full_dataset, x_mean=train_mean)
```

There's no `_compute_x_mean_from`. The actual method is `_compute_x_mean(self, dataset, feature_key)` — instance method, different name. A user copying this example would get an `AttributeError`. Either rename `_compute_x_mean` to `_compute_x_mean_from` and make it a `@staticmethod`, or update the docstring example to match the actual API.

Three small nits still open from my last pass (none are blockers, all are test additions):

1. `test_model_forward` doesn't assert `"logit" in out` even though `forward` now returns it.
2. No test exercises `GRUD.prepare_input`. The round-trip relationship with `_split_channels` documented in the docstring isn't actually verified anywhere.
3. `test_filter_matrix_not_learnable` only checks `requires_grad=False`, which would also pass for the old `nn.Parameter(..., requires_grad=False)` form. Mirror `test_x_mean_is_buffer` to actually verify the buffer change.

Also a tiny notebook issue: there are two Section 7 headers ("Ablation 2 — Decay Mechanism" and "Decay Mechanism on MIMIC-III Data"). Bump the second one to Section 8.

The `x_mean` indexing is the one I'd most like resolved before merge. The rest can land here or in a follow-up.

examples/mortality_prediction/mimic3_mortality_grud.ipynb
jhnwu3, line 1: This is the correct way of quickly porting over a dataset to PyHealth without doing any of the previous engineering steps, but length of stay and MIMIC3 are already supported within PyHealth here. 

Making a comment such that if we ever come back to this PR, it's pretty much there. Just one more task file and we have an end to end replication here.

PR #994, opened 2026-04-15
Final project submission: temporal MIMIC3 pipeline

elizabethbinkina:
Name: Elizabeth Binkina
NetID: binkina2

Type of Contribution:
Full pipeline (Dataset + Task + Model)

Paper:
https://proceedings.mlr.press/v106/nestor19a.html

Description:
This PR introduces a temporal extension of the MIMIC-III dataset, along with a hospital mortality prediction task and a neural network model that incorporates temporal features.

The goal of this contribution is to enable modeling of temporal trends in patient data by incorporating admission year as a feature and supporting temporally consistent dataset splits.

Key features:
- Temporal normalization of admission year (scaled between 0 and 1)
- Temporal train/validation/test splits based on admission year
- Integration of temporal features into downstream prediction tasks
- Lightweight neural model (TemporalFusionMLP) for combining temporal and clinical features

Files Added/Modified:
- pyhealth/datasets/mimic3_temporal.py
- pyhealth/datasets/configs/mimic3_temporal.yaml
- pyhealth/tasks/temporal_hospital_mortality_prediction.py
- pyhealth/models/temporal_fusion_mlp.py
- examples/mimic3_temporal_hospital_mortality_temporalfusionmlp.py
- tests/core/test_mimic3_temporal.py
- tests/core/test_temporal_fusion_mlp.py
- tests/core/test_temporal_hospital_mortality_prediction.py
- docs/api/datasets/pyhealth.datasets.mimic3_temporal.rst
- docs/api/models/pyhealth.models.temporal_fusion_mlp.rst
- docs/api/tasks/pyhealth.tasks.temporal_hospital_mortality_prediction.rst

Example Usage:
Run the following script to reproduce results:

python examples/mimic3_temporal_hospital_mortality_temporalfusionmlp.py

This script demonstrates:
- Random vs temporal dataset splits
- Model performance with and without temporal features

Testing:
Unit tests are included for:
- Dataset utilities (normalization and temporal splitting)
- Task sample generation and modality requirements
- Model instantiation, forward pass, and gradient computation

All tests use synthetic data and run quickly.


joshuasteier:
Hi @elizabethbinkina, thank you for the detailed pipeline submission. 

Before I can approve the PR, there are a few things I need you to address, and I want to flag the most important one first because I suspect it was unintentional.

**Unintended deletions in `pyhealth/datasets/__init__.py` and `pyhealth/models/__init__.py`**

Your diff shows `+10 -90` on `pyhealth/datasets/__init__.py` and `+7 -44` on `pyhealth/models/__init__.py`. After your PR, the datasets `__all__` would contain only `BaseDataset`, `SampleDataset`, `SampleBuilder`, `MIMIC3Dataset`, and your new `MIMIC3TemporalDataset`. The models `__all__` would contain only `BaseModel`, `EmbeddingModel`, and `TemporalFusionMLP`. Everything else including `MIMIC4Dataset`, `eICUDataset`, the signal and imaging datasets, most splitter and collate utilities, and the full model zoo would be removed from the public API.

This looks like your local `__init__.py` got committed in place of adding to the existing one. Could you rebase on `master` and add your new imports without removing anything else? If you open these two files on the current `master` branch and diff against what you have, you should see what needs to be preserved.

**`tests/conftest.py` installs a fake `litdata` module**

Your `conftest.py` creates a stub `litdata` module and registers it in `sys.modules` via `_install_fake_litdata()` before any test runs. Because conftest files apply to the whole test suite, this stub replaces the real `litdata` for every other test in the repo, not just yours. It also pairs with the try/except guards you added to `pyhealth/datasets/sample_dataset.py` and `pyhealth/datasets/utils.py`, which silently fall back when `litdata` is not available.

Could you remove the `conftest.py` stub and the import guards? If the real `litdata` is missing in your environment, the fix is to add it to the dev requirements or your test environment, not to stub it globally. If your tests genuinely do not need litdata, the fixtures can mock only what they need, scoped to your own test files.

**Duplicate empty test files at `tests/` root**

The real tests are at `tests/core/test_mimic3_temporal.py`, `tests/core/test_temporal_fusion_mlp.py`, and `tests/core/test_temporal_hospital_mortality_prediction.py`. The diff also adds three whitespace-only files with the same names at the root of `tests/`. Please delete the empty ones.

**`TemporalFusionMLP.forward` signature does not match `BaseModel`**

Other PyHealth models such as `LogisticRegression` accept the batch dict from the dataloader via `forward(**kwargs)` and return `{"loss", "y_prob", "y_true", "logit"}` so they work with `pyhealth.trainer.Trainer`. Your model takes a raw tensor and returns logits. This means the trainer cannot drive it, and your unit tests pass because they call `model(x)` directly with a manually constructed tensor.

Could you rework `forward` to:

1. Accept `**kwargs` (the batch dict),
2. Build the input tensor from the feature keys using the embedding pattern from `LogisticRegression`,
3. Compute the loss internally using `self.get_loss_function()`,
4. Return the standard dict with `loss`, `y_prob`, `y_true`, and `logit`.

Along the same lines, `BaseModel.__init__` expects a `dataset` argument and uses it to populate `self.feature_keys` and `self.label_keys`. Please pass the dataset explicitly and make sure the model reads from `self.feature_keys` rather than assuming a flat `input_dim`. The `DummyDataset` stub in your test would then need to look more like a real `SampleDataset`.

**Smaller items**

1. `_normalize_year` is implemented once in `MIMIC3TemporalDataset` and again in `TemporalHospitalMortalityPredictionMIMIC3`. Please have the task import it from the dataset module or share a helper, so the canonical clipping range stays consistent.

2. `TemporalSplit` is a nice touch. It would be worth returning this from a convenience method on the dataset that pulls `admission_year_raw` from its own samples, so users do not have to assemble the years list themselves.

3. The example script is 433 lines. If there is a way to split it into a short runnable demo plus a longer ablation script, that would make the example easier for new users to pick up.

4. Small typo: the docstring header of `pyhealth/models/temporal_fusion_mlp.py` reads `"Author:Elizabeth Binkina"` (missing space after the colon).


Once the `__init__.py` issue and the conftest stub are resolved, the rest should be relatively straightforward. The core idea is a good contribution to PyHealth. I just want to make sure we do not accidentally remove other parts of the package in the process.

elizabethbinkina:
Thank you for the detailed feedback — I’ve now addressed the requested changes.

* restored `pyhealth/datasets/__init__.py` and `pyhealth/models/__init__.py` so the existing public API is preserved while only adding the new dataset/model exports
* removed the global `litdata` stub and reverted the fallback import guards
* deleted the duplicate empty test files at the root of `tests/`
* reworked `TemporalFusionMLP` to align with the `BaseModel` interface used in this repository
* removed duplicated year-normalization logic from the task and reused the dataset normalization helper
* fixed the remaining import issues affecting local test collection

I also re-ran the relevant tests locally:

* temporal dataset tests passed
* temporal task tests passed
* temporal model tests passed

PR #999, opened 2026-04-16
CaliForest - kobeguo2

KobeGuo99:
**Contributor:** Kobe Guo (kobeguo2)

**Contribution Type:** Model

**Description:**
This PR implements CaliForest, a calibrated Random Forest model for healthcare prediction tasks. The model uses out-of-bag (OOB) predictions and tree-level variance to perform post-hoc probability calibration via isotonic regression or logistic regression, improving probability reliability.

The implementation integrates with PyHealth’s BaseModel API and supports PyHealth training and evaluation workflows.

**Files to Review:**
- `pyhealth/models/califorest.py` — core model implementation
- `tests/core/test_califorest.py` — unit tests using synthetic data
- `examples/mimic4_califorest.py` — real-data ablation on MIMIC-IV (RF vs CaliForest, isotonic vs logistic)
- `docs/api/models/pyhealth.models.califorest.rst` — API documentation
- `docs/api/models.rst` — updated model index
- `pyhealth/models/__init__.py` — model registration

**Testing:**
- All unit tests pass locally (`pytest tests/core/test_califorest.py`)
- Example ablation results (MIMIC-IV subset):
  - RF AUROC: ~0.80
  - CaliForest (isotonic): ~0.81
  - CaliForest (logistic): ~0.80

**Notes:**
- Dataset path is configurable via `MIMIC4_ROOT` environment variable
- Synthetic data is used in tests to ensure fast CI execution

### DIFF
--- a/docs/api/models.rst
+++ b/docs/api/models.rst
@@ -204,3 +204,4 @@ API Reference
     models/pyhealth.models.TextEmbedding
     models/pyhealth.models.BIOT
     models/pyhealth.models.unified_multimodal_embedding_docs
+    models/pyhealth.models.califorest
--- a/docs/api/models/pyhealth.models.califorest.rst
+++ b/docs/api/models/pyhealth.models.califorest.rst
@@ -1,0 +1,7 @@
+pyhealth.models.califorest
+==========================
+
+.. automodule:: pyhealth.models.califorest
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/examples/mimic4_califorest.py
+++ b/examples/mimic4_califorest.py
@@ -81,7 +81,6 @@ def run_califorest(
         test_dataset, batch_size=len(test_dataset), shuffle=False
     )
 
-    train_batch = next(iter(train_loader))
     test_batch = next(iter(test_loader))
 
     model = CaliForest(
@@ -90,8 +89,7 @@ def run_califorest(
         calibration=calibration,
         random_state=42,
     )
-
-    _ = model(**train_batch)
+    model.fit(train_loader)
 
     with torch.no_grad():
         ret = model(**test_batch)
--- a/examples/mimic4_califorest.py
+++ b/examples/mimic4_califorest.py
@@ -1,0 +1,192 @@
+from __future__ import annotations
+
+import os
+
+import numpy as np
+import torch
+from sklearn.ensemble import RandomForestClassifier
+from sklearn.metrics import brier_score_loss, roc_auc_score
+
+from pyhealth.datasets import (
+    MIMIC4EHRDataset,
+    create_sample_dataset,
+    get_dataloader,
+)
+from pyhealth.models import CaliForest
+from pyhealth.tasks import InHospitalMortalityMIMIC4
+
+
+# Set your MIMIC-IV dataset path via environment variable before running:
+# export MIMIC4_ROOT=/your/path/to/mimiciv/3.1
+ROOT = os.getenv("MIMIC4_ROOT")
+
+
+def evaluate(y_true: np.ndarray, y_prob: np.ndarray) -> dict[str, float]:
+    """Compute AUROC and Brier score."""
+    y_true = np.asarray(y_true).reshape(-1)
+    y_prob = np.asarray(y_prob).reshape(-1)
+    return {
+        "auroc": float(roc_auc_score(y_true, y_prob)),
+        "brier": float(brier_score_loss(y_true, y_prob)),
+    }
+
+
+def run_califorest(
+    X_train: np.ndarray,
+    y_train: np.ndarray,
+    X_test: np.ndarray,
+    y_test: np.ndarray,
+    calibration: str,
+) -> dict[str, float]:
+    """Train and evaluate CaliForest on tabularized features."""
+    train_samples = []
+    for i in range(len(X_train)):
+        train_samples.append(
+            {
+                "patient_id": f"train-{i}",
+                "visit_id": f"train-{i}",
+                "features": X_train[i].tolist(),
+                "label": int(y_train[i]),
+            }
+        )
+
+    test_samples = []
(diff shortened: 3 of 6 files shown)


joshuasteier:
Hi @KobeGuo99 , thanks for putting this together. The BaseModel integration, SampleDataset handling, the docs entry, and the model registration in __init__.py all look correct and match the pattern used by LogisticRegression in the package. 

The MIMIC-IV ablation in examples/mimic4_califorest.py is a nice touch and gives a real healthcare-relevant comparison.I have one thing I would like you to think through, and a few smaller suggestions.

I have one thing I would like you to think through, and a few smaller suggestions.
The main thing is around when the random forest actually gets trained. In forward:

if not self.is_fitted:
    self.fit_model(**kwargs)
    
    
This fits the RF and the calibrator on the first batch seen during forward, then freezes them for the rest of training. If someone uses the standard pyhealth.trainer.Trainer with normal mini-batching, the RF ends up trained on just that first batch and never sees the rest of the data. The test_model_backward test still passes because self.logit_scale has a gradient, but the forest itself is not being updated across epochs.Your MIMIC-IV example works around this by using batch_size=len(train_dataset), which feeds the whole training set in one batch. That is fine for the example, but it would be good to make this behavior explicit so future users do not hit it by surprise. A few ways you could handle it:
Add a clear note in the class docstring and in fit_model that CaliForest expects the full training set in one batch, and add an assertion or warning when it looks like mini-batching is happening.
Add a separate fit(train_dataset) method that materializes the full dataset once and fits the RF there. Then forward only does inference. This would decouple the RF from the torch training loop and make the behavior easier to reason about.
Keep forward as is but add a short section to the class docstring explaining the design choice and pointing users to the example.


Option 2 is probably the cleanest fit for how tree models usually work, but option 1 or 3 is much less work if you are short on time .Smaller suggestions:
The tests verify shapes and that a gradient flows, but they do not check the paper's main claim that CaliForest improves calibration over a vanilla random forest. It would strengthen the PR to add one test that compares Brier score between CaliForest and a plain RandomForestClassifier on synthetic data with maybe 200 to 500 samples, and asserts CaliForest is not worse. Also a quick test that isotonic and logistic produce different outputs on the same input would confirm both code paths are live.



The PR description mentions tree-level variance weighting, and I can see it in fit_model as weights = 1.0 / (variances + 1e-6). A short comment in the code pointing to the paper section that motivates this, and maybe a test that confirms the weighting changes the calibrator fit compared to unweighted, would make this feature easier for reviewers and users to verify.

predict_proba_numpy uses [:, 1] indexing which assumes binary classification. Could you add an assertion in fit_model that the labels are binary, or mention the binary-only scope in the class docstring? That way users with a multiclass task get a clear error instead of a silent wrong answer.

The unit tests use 4 samples with n_estimators=10. That is on the small side for OOB to be meaningful. Bumping to 50 or 100 synthetic samples would make the tests more representative.

One small thing: there is a parallel PR at #995 from a different team that also implements CaliForest. If you two want to compare notes, their implementation includes a min_oob_trees reliability threshold that is worth looking at for a possible follow-up addition. No action needed here, just a heads up.
Overall the structure is solid and the integration with PyHealth is done right. Happy to iterate on any of the above.

KobeGuo99:
Thank you! This was a good catch. I updated CaliForest so it now requires an explicit `fit(train_loader)` call and no longer trains inside `forward()`. `forward()` now raises a clear error if called before fitting. I also updated the MIMIC-IV example to call `model.fit(train_loader)` explicitly before inference, added a binary-only note in the class docstring, and updated the tests accordingly. I also added a small test showing that the isotonic and logistic calibration paths produce different outputs. I really appreciate the feedback!

joshuasteier:
Thanks @KobeGuo99 , this is a clean response to the feedback. The explicit `fit(train_loader)` method decouples the RF fitting from the torch training loop, the `forward` error message when called before fit is clear, the binary-label check in `fit_model` gives users a clear failure mode, and `test_isotonic_and_logistic_differ` confirms both calibration paths are active. The docstring note about binary scope and the fit-before-inference contract is exactly what I wanted.

A few small things worth thinking about, none blocking:

1. `self.logit_scale = nn.Parameter(torch.tensor(1.0))` is still defined and multiplied into the logits in `forward`. Under the old design where `forward` also did the fit, `logit_scale` gave `Trainer.fit()` something to update. Under the new design where training happens in `model.fit(train_loader)` and `forward` is inference-only, `logit_scale` sits as an unused trainable parameter. You can either remove it (simplest), or keep it and wire up a small optimiser step inside `fit(train_loader)` that tunes it on the train set (a Platt-scaling-on-top-of-calibration step). Either is fine; I would just not leave it dangling.

2. The unit tests still use 4 samples with `n_estimators=10`. OOB with so few samples is noisy. Bumping to 50 or 100 synthetic samples would make the tests more representative, and would also enable a quick Brier-score comparison test between CaliForest and a plain `RandomForestClassifier` to verify the paper's calibration claim. Not required for this PR, but worth it as a follow-up.

3. The comment on variance weighting (`CaliForest uses inverse tree-level variance so more stable predictions have greater influence during calibrator fitting`) explains the why, which is what I wanted. If you have the paper handy, a one-liner pointer to the equation or section would make it even easier for future reviewers to verify.

Happy to approve once you either remove or wire up `logit_scale`. The test additions are nice-to-have and can land in a follow-up PR.

KobeGuo99:
Thank you! I removed the logit_scale from califorest.py and reran the CaliForest tests and the MIMIC-IV example after the change.

joshuasteier (approved):
Thanks @KobeGuo99.

One tiny nit, not blocking: `import torch.nn as nn` at the top of the file is now unused since `nn.Parameter` was the only thing using it. Feel free to remove the import on your next push, or leave it and a follow-up cleanup will catch it.

Approving. LGTM.

PR #1003, opened 2026-04-17
[DL4H Sp26] Add MedFuse model for multi-modal EHR + chest X-ray fusion

seannian:
Contributor: Sean Nian (snian2@illinois.edu), Tony Hong (tonyhong@illinois.edu), Yaqi Qiao (yaqiq2@illinois.edu)
Contribution Type: Model
Paper: https://arxiv.org/abs/2207.07027

Description
Add MedFuse, a multimodal clinical prediction model that fuses EHR time-series and chest X-ray representations using LSTM-based sequential fusion, with support for missing CXR modality.

Files to review
pyhealth/models/medfuse.py - New MedFuse model implementation including:

> MedFuse - Main PyHealth-integrated BaseModel subclass
> MedFuseLayer - Fusion layer with EHR LSTM encoder, ResNet CXR encoder, projection, and fusion LSTM

pyhealth/models/init.py - Added imports for MedFuse and MedFuseLayer
tests/core/test_medfuse.py - Unit tests with synthetic data covering instantiation, forward pass, output shape, gradient flow, missing modality, and binary/multilabel modes
examples/mimic4_mortality_medfuse.py - Ablation study varying EHR hidden dim, fusion hidden dim, dropout, learning rate, and modality
docs/api/models/pyhealth.models.MedFuse.rst - Sphinx documentation page
docs/api/models.rst - Added MedFuse entry to models table of contents

### DIFF
--- a/docs/api/models.rst
+++ b/docs/api/models.rst
@@ -194,6 +194,7 @@ API Reference
     models/pyhealth.models.ConCare
     models/pyhealth.models.Agent
     models/pyhealth.models.GRASP
+    models/pyhealth.models.MedFuse
     models/pyhealth.models.MedLink
     models/pyhealth.models.TCN
     models/pyhealth.models.TFMTokenizer
--- a/docs/api/models/pyhealth.models.MedFuse.rst
+++ b/docs/api/models/pyhealth.models.MedFuse.rst
@@ -1,0 +1,24 @@
+pyhealth.models.MedFuse
+=======================
+
+Overview
+--------
+
+MedFuse is an LSTM-based multi-modal fusion model for combining clinical
+time-series (EHR) and chest X-ray (CXR) representations.
+
+Reference: Hayat et al., "MedFuse: Multi-modal fusion with clinical
+time-series data and chest X-ray images." MLHC 2022.
+
+API Reference
+-------------
+
+.. autoclass:: pyhealth.models.MedFuseLayer
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+.. autoclass:: pyhealth.models.MedFuse
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/examples/mimic4_mortality_medfuse.py
+++ b/examples/mimic4_mortality_medfuse.py
@@ -1,0 +1,381 @@
+"""MedFuse Ablation Study for In-Hospital Mortality Prediction.
+
+This script demonstrates MedFuse with varying hyperparameters on synthetic
+mortality-like data to validate model behavior quickly.
+
+Paper:
+    Hayat et al. "MedFuse: Multi-modal fusion with clinical time-series data
+    and chest X-ray images." MLHC 2022.
+
+Ablations:
+    1. EHR hidden dim: 64, 128, 256, 512
+    2. Fusion hidden dim: 128, 256, 512
+    3. Dropout: 0.0, 0.3, 0.5, 0.7
+    4. Learning rate: 1e-5, 1e-4, 1e-3
+    5. Modality: EHR-only vs EHR+CXR
+
+The script is intentionally lightweight by default:
+    - Synthetic data only
+    - Small dataset
+    - Few epochs
+"""
+
+from __future__ import annotations
+
+import argparse
+import math
+import random
+from typing import Dict, List, Tuple
+
+import numpy as np
+import torch
+from sklearn.metrics import average_precision_score, roc_auc_score
+
+from pyhealth.datasets import create_sample_dataset, get_dataloader
+from pyhealth.models import MedFuse
+
+
+def set_seed(seed: int) -> None:
+    """Sets all random seeds for reproducibility."""
+    random.seed(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+
+
+def make_synthetic_samples(
+    num_samples: int,
+    seq_len: int,
+    ehr_dim: int,
+    image_size: int,
+    seed: int,
+) -> List[Dict[str, object]]:
+    """Builds synthetic samples with weakly learnable mortality signal."""
+    generator = torch.Generator().manual_seed(seed)
+    samples: List[Dict[str, object]] = []
+
+    for index in range(num_samples):
+        ehr = torch.randn(seq_len, ehr_dim, generator=generator)
+        cxr = torch.randn(3, image_size, image_size, generator=generator)
+
+        score = 0.7 * float(ehr[:, 0].mean())
+        score += 0.3 * float(cxr.mean())
+        score += float(torch.randn(1, generator=generator)) * 0.1
+        label = 1 if score > 0 else 0
+
+        samples.append(
+            {
+                "patient_id": f"patient-{index}",
+                "visit_id": f"visit-{index}",
+                "ehr": ehr.tolist(),
+                "cxr": cxr.tolist(),
+                "label": label,
+            }
+        )
+
+    return samples
+
+
+def build_datasets(
+    train_size: int,
+    val_size: int,
+    seq_len: int,
+    ehr_dim: int,
+    image_size: int,
+    seed: int,
+):
+    """Creates synthetic train/validation datasets."""
+    train_samples = make_synthetic_samples(
+        num_samples=train_size,
+        seq_len=seq_len,
+        ehr_dim=ehr_dim,
+        image_size=image_size,
+        seed=seed,
+    )
+    val_samples = make_synthetic_samples(
+        num_samples=val_size,
+        seq_len=seq_len,
+        ehr_dim=ehr_dim,
+        image_size=image_size,
+        seed=seed + 1,
+    )
+
+    input_schema = {"ehr": "tensor", "cxr": "tensor"}
+    output_schema = {"label": "binary"}
+
+    train_dataset = create_sample_dataset(
+        samples=train_samples,
+        input_schema=input_schema,
+        output_schema=output_schema,
+        dataset_name="medfuse_ablation_train",
+    )
+    val_dataset = create_sample_dataset(
+        samples=val_samples,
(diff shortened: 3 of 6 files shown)


joshuasteier:
Hi Sean(@seannian) (and Tony, Yaqi), this is a well-structured PR. The `BaseModel` and `MedFuseLayer` split mirrors the repo convention (compare `RETAIN`/`RETAINLayer`, `ConCare`/`ConCareLayer`), the integration with `SampleDataset`, `get_output_size`, `prepare_y_prob`, and `get_loss_function` is clean, and the sequential-fusion handling of missing CXR via `pack_padded_sequence` with per-sample sequence lengths is the correct way to do it. A few things worth addressing.

**Verify the fusion architecture against the paper.**

Your fusion layer treats the EHR and CXR representations as tokens in a length-1 or length-2 sequence fed to a fusion LSTM. I believe this matches the reference MedFuse implementation, but the paper also describes a variant where the EHR LSTM's hidden state initializes the fusion LSTM. Could you add a comment in `MedFuseLayer` pointing to the specific paper section (equation number or figure) that this implementation follows? That will save future reviewers from having to cross-check against the reference code.

**`_resolve_mode` reference looks inherited but worth confirming.**

In `MedFuse.__init__`:

```python
self.mode = self._resolve_mode(self.dataset.output_schema[self.label_key])
```

`_resolve_mode` is not defined in this file. I assume it is inherited from `BaseModel`. Can you confirm? If yes, ignore. If no, the model will fail at instantiation and the tests must be passing for a different reason.

**The ablation example runs but does not report numbers.**

The example defines a clean ablation over `ehr_hidden_dim`, `fusion_hidden_dim`, `dropout`, `lr`, and `modality`, but the PR does not include any output. With `train_size=32, epochs=1`, even the synthetic numbers would be noisy, but it would still be evidence that the model responds to hyperparameters and that EHR+CXR behaves differently from EHR-only. Could you run the example once and either paste the resulting table into the PR description, or commit a run log alongside the script? Right now the reproducibility claim rests on tests rather than ablation results.

**`cxr_mask` is accepted as a kwarg but is not in the test dataset's input schema.**

In `test_missing_modality_robustness`, you construct `cxr_mask` manually and pass it as a kwarg directly to `model(...)`. The standard PyHealth dataloader built from `input_schema={"ehr": "tensor", "cxr": "tensor"}` will not produce a `cxr_mask` key automatically. Users who want the mask to flow through the normal `dataloader -> model(**batch)` path need to include `cxr_mask` in their dataset's `input_schema`.

Could you add a note to the `MedFuse` docstring (in the `cxr_mask_key` argument description) explaining how users should surface the mask through their dataset? Even a one-line example like "To use `cxr_mask`, include `'cxr_mask': 'tensor'` in your input schema with per-sample availability values" would help.

**`test_gradient_computation` references a phantom `_dummy_param`.**

```python
for name, parameter in model.named_parameters():
    if name == "_dummy_param":
        continue
    ...
    if parameter.grad is None:
        print(f"Gradient missing for: {name}")
    self.assertIsNotNone(parameter.grad)
```

There is no `_dummy_param` in the `MedFuse` or `MedFuseLayer` code I can see. If this is a leftover from development, please remove the exception. The `print` inside the test should also be removed (it will spam test output in CI). If `_dummy_param` is added by `BaseModel`, a comment saying so would help future readers.

**CXR encoder is built even when training EHR-only.**

In the modality ablation, the EHR-only configuration still instantiates the ResNet backbone inside `MedFuseLayer.__init__`. The parameters exist but receive no gradients. Not a correctness bug, but wasteful. If you expect this to be a common use case, consider making the CXR encoder construction conditional on whether the dataset has CXR in its feature keys. If it is meant to always be present, ignore.

**Smaller items**

1. PR description lists three contributors but the file header says only "Sean Nian." Please list all three in the file header for consistent attribution.

2. `cast(torch.Tensor | tuple[torch.Tensor, ...], ...)` uses PEP 604 union syntax that `cast()` evaluates at runtime, so `from __future__ import annotations` does not help here. This requires Python 3.10+. Could you confirm PyHealth's minimum Python version? If it is 3.9 or lower, swap to `Union[torch.Tensor, Tuple[torch.Tensor, ...]]` from `typing`.

3. `_infer_ehr_input_dim` iterates `self.dataset` with a `for` loop. It returns on the first valid sample, so it is O(1) in practice, but `self.dataset[0]` or `next(iter(self.dataset))` would be more explicit. Very minor.

4. The example's closing note says "the paper reports that multi-modal EHR+CXR fusion improves over unimodal baselines." Please cite the specific numbers from Hayat et al. (2022) (for example, the mortality AUROC numbers from Table 2 or 3) so the claim is verifiable.

5. The test mock replaces `cxr_encoder` with a small synthetic network for speed. That is a good choice for CI. Worth adding a separate slower test gated on an env variable (for example `@unittest.skipUnless(os.environ.get("RUN_SLOW"))`) that uses the real ResNet at least once, to catch regressions in the real backbone integration.

seannian:
1. Checked the paper (§3.2, Eqs. 3–5, Fig. 2) — the fusion LSTM consumes [v_ehr, v*_cxr] as a token sequence, EHR is the first input token, not an initial hidden state. No such variant exists. Same in the reference code: https://github.com/nyuad-cai/MedFuse/blob/main/models/fusion.py — forward_lstm_fused calls self.lstm_fusion_layer(feats) with no (h_0, c_0), so it's zero-initialized. Added a pointer to §3.2, Eq. 4 in the MedFuseLayer docstring.
2. Inherited from BaseModel (base_model.py:80). No action needed.
3. See below
4. Added a note to the cxr_mask_key arg docstring explaining that users need to include "cxr_mask": "tensor" in their input_schema to have the mask flow through the standard dataloader.
5. Fixed — _dummy_param is a 0-element device-query helper from BaseModel (base_model.py:55), so the skip is correct. Removed the stray print(...) and added a comment explaining why it's skipped.
6. This is intentional and matches the paper (§3.2) — the CXR encoder is always instantiated; the modality flag only gates what gets fed into the forward pass. 

Smaller items:
1. Added Co-Authors
2. The code compiles and runs fine in CI — PyHealth requires Python >=3.12 (pyproject.toml:23), so PEP 604 X | Y at runtime should be supported.
3. Leaving as-is since it returns on the first valid sample.
4. Added the specific numbers from Hayat et al. (2022)
5. Going to skip this one; the class rubric explicitly requires fast tests (-3 pts for slow tests, milliseconds per test), so the mocked CXR encoder is intentional.

Ran `python examples/mimic4_mortality_medfuse.py` with defaults (`--train-size 32 --val-size 16 --epochs 1 --cxr-backbone resnet18`, `cxr_pretrained=False`) on an A100.

<details>
<summary>Full ablation log</summary>

```
EHR Hidden Dim Ablation
| setting | AUROC | AUPRC |
|---------|-------|-------|
| ehr_hidden_dim=64  | 0.2917 | 0.2202 |
| ehr_hidden_dim=128 | 0.8750 | 0.6917 |
| ehr_hidden_dim=256 | 0.7917 | 0.6250 |
| ehr_hidden_dim=512 | 0.4583 | 0.2932 |

Fusion Hidden Dim Ablation
| setting | AUROC | AUPRC |
|---------|-------|-------|
| fusion_hidden_dim=128 | 0.5000 | 0.2833 |
| fusion_hidden_dim=256 | 0.5833 | 0.5146 |
| fusion_hidden_dim=512 | 0.4375 | 0.2645 |

Dropout Ablation
| setting | AUROC | AUPRC |
|---------|-------|-------|
| dropout=0.0 | 0.4792 | 0.3702 |
| dropout=0.3 | 0.4167 | 0.2554 |
| dropout=0.5 | 0.6667 | 0.5271 |
| dropout=0.7 | 0.5833 | 0.3631 |

Learning Rate Ablation
| setting | AUROC | AUPRC |
|---------|-------|-------|
| learning_rate=1e-05 | 0.6667 | 0.4667 |
| learning_rate=1e-04 | 0.7292 | 0.6841 |
| learning_rate=1e-03 | 0.1458 | 0.1857 |

Modality Ablation
| setting | AUROC | AUPRC |
|---------|-------|-------|
| EHR-only | 0.7708 | 0.4512 |
| EHR+CXR  | 0.4583 | 0.2798 |

Findings:
- EHR Hidden Dim: best ehr_hidden_dim=128 (AUROC=0.8750)
- Fusion Hidden Dim: best fusion_hidden_dim=256 (AUROC=0.5833)
- Dropout: best dropout=0.5 (AUROC=0.6667)
- Learning Rate: best learning_rate=1e-04 (AUROC=0.7292)
- Modality: best EHR-only (AUROC=0.7708)
```

</details>

Numbers are noisy as expected, but the model clearly responds to hyperparameters (AUROC 0.15–0.88 across configs, LR=1e-3 collapses) and EHR+CXR behaves differently from EHR-only (0.458 vs 0.771).

joshuasteier (approved):
LGTM, closing this out. Thank you, @seannian.

seannian:
Hi @joshuasteier; sorry a bit new to this, seems like it needs a workflow run to be approved before it can get merged?

tonyhong007:
Hi @joshuasteier,

Just wanted to quickly follow up on the workflow approval.

tonyhong007:
Hi @joshuasteier,

Could we get the workflow approval?

joshuasteier:
@tonyhong007, sorry about that, it's approved and LGTM.

tonyhong007:
Hi @joshuasteier,

Thank you for approving the workflow. I believe we can't merge the PR in from our end. Is it possible for you to merge it in for us?

Jathurshan0330 (approved):
PR looks good.

tonyhong007:
@joshuasteier @Jathurshan0330 

Could you please help us merge this into main?

Rian354:
LGTM, Merging, thanks for your hard work!

PR #1138, opened 2026-04-25
Fix drug recommandation drug code and padding. 

### LINKED ISSUE #1137: PyHealth 2.0 cannot reproduce previous MIMIC-III drug recommendation / SafeDrug performance
中文：

我在 PyHealth 2.0 中发现两个问题，它们叠加后会导致旧版 MIMIC-III 药物推荐 / SafeDrug 结果难以复现。

第一，内置 DrugRecommendationMIMIC3 在 MIMIC-III 上使用的是 prescriptions/drug 文本字段，并直接截取前 4 个字符。这个输出既不是 NDC，也不是 ATC，更没有被转换到 ATC level 3。这一点与 1.x 版本的 PyHealth 不一致。

第二，EmbeddingModel 对 NestedSequenceProcessor 使用的是 padding_idx=None。这对 SafeDrug 不合适，因为模型会在 visit 内对 code embedding 直接求和。如果 padding embedding 是可训练的非零向量，那么 visit 表示就会被 padding token 污染。对于 SafeDrug 这种依赖 sum pooling 的结构，这种影响会非常明显。我本地测试中，仅把这里改成 padding_idx=0，SafeDrug 性能就显著恢复，并明显接近旧版本结果。

因此，即使本地修复了药物标签生成问题，PyHealth 2.0 仍然无法直接复现旧版药物推荐性能，除非同时修改 padding 行为。

建议：

修改 DrugRecommendationMIMIC3，让 MIMIC-III 的药物标签来自与 SafeDrug 兼容的正式药物编码体系，而不是截断后的自由文本药名。
对 SafeDrug 这类会对 code embedding 做 sum pooling 的模型，使用零 padding embedding（例如 padding_idx=0），或者在求和前显式 mask 掉 padding token。
English:

I found two issues in PyHealth 2.0 that together make previous MIMIC-III drug recommendation / SafeDrug results difficult to reproduce.

First, the built-in DrugRecommendationMIMIC3 task for MIMIC-III uses the prescriptions/drug text field and directly truncates it to the first 4 characters. This output is neither NDC nor ATC, and it is not converted to ATC level 3. This behavior is inconsistent with PyHealth 1.x. 

Second, EmbeddingModel uses padding_idx=None for NestedSequenceProcessor. This is problematic for SafeDrug, because the model directly sums code embeddings within each visit. If the padding embedding is trainable and non-zero, visit representations become contaminated by padding tokens. This is especially harmful for models like SafeDrug that rely on sum pooling. In my local tests, simply changing this to padding_idx=0 significantly restores SafeDrug performance and makes it much closer to the previous version.

Therefore, even after locally fixing the drug label generation issue, PyHealth 2.0 still cannot directly reproduce previous drug recommendation performance unless the padding behavior is also changed.

Suggestions:

Update DrugRecommendationMIMIC3 so that MIMIC-III drug labels come from a proper medication coding system compatible with SafeDrug, instead of truncated free-text drug names.
For models like SafeDrug that perform sum pooling over code embeddings, use zero padding embeddings (for example padding_idx=0), or explicitly mask out padding tokens before summation.

Logiquo:
Fix https://github.com/sunlabuiuc/PyHealth/issues/1137

### DIFF
--- a/pyhealth/tasks/drug_recommendation.py
+++ b/pyhealth/tasks/drug_recommendation.py
@@ -1,9 +1,55 @@
-from typing import Any, Dict, List
+from typing import Any, Dict, Iterable, List, Optional
 
 import polars as pl
 
 from pyhealth.data import Patient, Visit
+from pyhealth.medcode import CrossMap
 from .base_task import BaseTask
+
+
+_NDC_TO_ATC3_MAPPER = None
+_NDC_TO_ATC3_CACHE: Dict[str, List[str]] = {}
+
+
+def _get_ndc_to_atc3_mapper():
+    global _NDC_TO_ATC3_MAPPER
+    if _NDC_TO_ATC3_MAPPER is None:
+        _NDC_TO_ATC3_MAPPER = CrossMap.load("NDC", "ATC")
+    return _NDC_TO_ATC3_MAPPER
+
+
+def _is_missing_ndc(code: Any) -> bool:
+    if code is None:
+        return True
+    code = str(code).strip()
+    return code == "" or code == "0" or code.lower() in {"nan", "none", "<na>"}
+
+
+def _map_ndc_list_to_atc3(
+    ndc_codes: Iterable[Any],
+    mapper: Optional[Any] = None,
+) -> List[str]:
+    """Maps MIMIC prescription NDCs to stable, deduplicated ATC-3 labels."""
+    mapper = _get_ndc_to_atc3_mapper() if mapper is None else mapper
+    drugs: List[str] = []
+    seen = set()
+
+    for ndc in ndc_codes:
+        if _is_missing_ndc(ndc):
+            continue
+        ndc = str(ndc).strip()
+        if ndc not in _NDC_TO_ATC3_CACHE:
+            _NDC_TO_ATC3_CACHE[ndc] = mapper.map(ndc, target_kwargs={"level": 3})
+        mapped_codes = _NDC_TO_ATC3_CACHE[ndc]
+        for code in mapped_codes:
+            if code is None:
+                continue
+            code = str(code).strip()
+            if code and code not in seen:
+                drugs.append(code)
+                seen.add(code)
+
+    return drugs
 
 
 class DrugRecommendationMIMIC3(BaseTask):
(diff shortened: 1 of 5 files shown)


.codex
jhnwu3, line 1: We should probably delete this .codex file before merging into the repo.

tests/core/test_drug_recommendation_atc3.py
```diff
@@ -0,0 +1,221 @@
...
+class FakeVisit:
+    def __init__(self, visit_id, table_codes):
+        self.visit_id = visit_id
+        self.table_codes = table_codes
+
+    def get_code_list(self, table):
+        return self.table_codes[table]
+
+
+class FakeLegacyPatient:
```
jhnwu3: curious what this is for? in the test casing?

I'm no longer opposed to this approach, but we should be able to generate some fake ndc samples using the demo dataset in test-resources/ if we really want to test it out on a fake dataset (ofc I guess that takes longer).

pyhealth/tasks/drug_recommendation.py
```diff
@@ -1,11 +1,57 @@
...
 from pyhealth.data import Patient, Visit
+from pyhealth.medcode import CrossMap
 from .base_task import BaseTask
 
 
+_NDC_TO_ATC3_MAPPER = None
+_NDC_TO_ATC3_CACHE: Dict[str, List[str]] = {}
+
+
+def _get_ndc_to_atc3_mapper():
```
jhnwu3, line 14: I forgot to mention, sequence processor can also handle the code mappings now too. Don't think it matters that much here, but worth sharing.

```
class SequenceProcessor(FeatureProcessor, TokenProcessorInterface):
    """Feature processor for encoding categorical sequences.

    Encodes medical codes (e.g., diagnoses, procedures) into numerical
    indices. Supports single or multiple tokens and can build vocabulary
    on the fly if not provided.

    Args:
        code_mapping: optional tuple of (source_vocabulary, target_vocabulary)
            to map raw codes to a grouped vocabulary before tokenizing.
            Uses ``pyhealth.medcode.CrossMap`` internally. For example,
            ``("ICD9CM", "CCSCM")`` maps ~128K ICD-9 diagnosis codes to
            ~280 CCS categories, and ``("NDC", "ATC")`` maps ~940K drug
            codes to ~5K ATC categories. When None (default), codes are
            used as-is with no change to existing behavior.

    Examples:
        >>> proc = SequenceProcessor()  # no mapping, same as before
        >>> proc = SequenceProcessor(code_mapping=("ICD9CM", "CCSCM"))
    """
```

jhnwu3 (changes requested):
tldr; Would be good to just delete the .codex file. 

I'm curious about the logic wit the test casing with the FakeLegacyPatient class here, but not too concerned at the moment.

pyhealth/tasks/drug_recommendation.py
Logiquo: This is a bit tricky because NestedSequence & MultiLabel does not yet support this. May be better to left this into a future PR.

jhnwu3 (approved):
new changes lgtm

PR #1148, opened 2026-05-18
Add synthetic-EHR generative evaluation metrics

chufangao:
## Summary
Adds a complete synthetic-EHR pipeline to PyHealth: a shared generation
**task**, five **generator models**, a generative **metrics** subpackage, an
end-to-end **example**, and tests for all of the above.

### 1. Shared task — `pyhealth/tasks/generate_ehr.py`
- `EHRGeneration` — base class that emits one sample per patient,
  `{"patient_id", "visits": [[code, ...], ...]}`, processed by
  `NestedSequenceProcessor`. Unconditional (no output labels).
- `EHRGenerationMIMIC3` / `EHRGenerationMIMIC4` — dataset-specific
  subclasses; only the event type and code attribute differ.
- Helpers `decode_dataset` and `to_evaluation_dataframe` convert real
  `SampleDataset`s and any generator's `generate()` output into the
  long-form `(id, time, visit_codes, labels)` dataframe that the metrics
  consume.

### 2. Generator models — `pyhealth/models/generators/`
All five share the `EHRGeneration` task and expose the same
`train_model(train_dataset, val_dataset=None)` / `generate(num_samples)` API.

| Model | File | Family |
|-------|------|--------|
| **HALO** | `halo.py` | Autoregressive multi-hot transformer (Theodorou et al., 2023) |
| **GPT2** | `gpt2.py` | Causal-LM sequence model over flattened code streams |
| **PromptEHR** | `promptehr.py` | BART denoising autoencoder with soft prompts (Wang & Sun, EMNLP'22) |
| **MedGAN** | `medgan.py` | Bag-of-codes VAE + GAN (Choi et al., MLHC'17) |
| **CorGAN** | `corgan.py` | Bag-of-codes correlation-aware GAN (Torfi & Fox, FLAIRS'20) |

Each model is registered as a sub-module so `.parameters()`/`.to(device)`
work, owns its own training loop (best-checkpoint saving included), and
returns decoded code strings from `generate()`.

### 3. Metrics subpackage — `pyhealth/metrics/generative/`
Evaluation along three axes:

- **`privacy.py`** — `calc_nnaar` (Nearest Neighbor Adversarial Accuracy
  Risk), `calc_membership_inference` (membership inference attack), and
  `compute_discriminator_privacy` (real-vs-synthetic discriminator score).
- **`utility.py`** — `compute_mle` (machine learning efficacy, TRTR vs TSTR)
  and `compute_prevalence_metrics` (code-prevalence similarity: R², Pearson,
  RMSE).
- **`utils.py`** — shared data prep, a self-contained LSTM classifier, and a
  random-forest baseline.
- **`evaluate_synthetic_ehr()`** — convenience orchestrator that runs the
  full suite and returns one merged `{metric: (mean, std)}` dict.

Metrics operate on flat dataframes (`id`, `time`, `visit_codes`, `labels`),
so they work for any generator. Public functions are re-exported from
`pyhealth.metrics`.

#### Port cleanups
- `logging` instead of bare `print` calls.
- Fixed a latent CUDA crash in the LSTM eval loop (`.cpu().numpy()`).
- Replaced `scipy.stats.pearsonr` with `numpy.corrcoef` to avoid an
  undeclared `scipy` dependency.
- Input dataframes are copied instead of mutated in place.
- Google-style docstrings, type hints, PEP8 (≤88 chars).

### 4. End-to-end example — `examples/halo_mimic3.py`
Loads MIMIC-III → applies `EHRGenerationMIMIC3` → trains HALO → generates
synthetic patients → runs the full `evaluate_synthetic_ehr` suite and prints
each `(mean, std)` metric. Verified to run end-to-end on the `dev=True`
subset (NNAAR, MIA, MLE TRTR/TSTR, discriminator privacy, prevalence all
produced).

### 5. Tests
All passing.

| File | Tests | Covers |
|------|-------|--------|
| `tests/core/test_generative_metrics.py` | 18 | per-metric + orchestrator + behavioral sanity |
| `tests/core/test_halo.py` | 7 | init, train, generate, encoding |
| `tests/core/test_gpt2.py` | 7 | init, train, generate |
| `tests/core/test_promptehr.py` | 7 | init, train, generate, prompt handling |
| `tests/core/test_medgan.py` | 9 | VAE pretrain + GAN train + generate |
| `tests/core/test_corgan.py` | 12 | conv autoencoder + GAN + generate |

The metrics suite includes 5 **behavioral tests** that verify each metric
responds sensibly across three synthetic datasets — an *exact copy* of the
training data, a *similar* set (~15% of codes perturbed), and a *different*
set (disjoint code vocabulary):

| Metric | Verified behavior |
|--------|-------------------|
| Prevalence | RMSE `0 → 0.03 → 0.26`; exact copy → RMSE 0, R²/Pearson = 1 |
| NNAAR | Flags memorization: `1.0 → 0.1 → 0.0` |
| Membership inference | Attack accuracy `1.0 → 0.94 → 0.46` (chance for unrelated data) |
| Discriminator privacy | Disjoint-vocabulary data trivially flagged; real-derived data is not |
| MLE (utility) | Exact copy reproduces real utility exactly; ratio degrades `1.0 → 0.98 → 0.81` |

## Notes
- MLE currently hard-codes next-visit prediction, which is degenerate for
  bag-of-codes generators (MedGAN, CorGAN) — those should be evaluated with
  `metrics="privacy"` plus the prevalence metrics. A future revision will
  let callers plug in a static-label task (mortality, readmission,
  "ever diagnosed with X") so MLE is meaningful for both families.
- The discriminator-privacy *score* is degenerate for exact copies (the
  model predicts a constant on identical features, so the score reflects
  test-split balance rather than 0.5). The behavioral test asserts the
  robust direction — disjoint synthetic data is cleanly flagged while
  real-derived data is not.

### DIFF
--- a/docs/api/metrics/pyhealth.metrics.generative.rst
+++ b/docs/api/metrics/pyhealth.metrics.generative.rst
@@ -1,0 +1,25 @@
+pyhealth.metrics.generative
+===================================
+
+Evaluation metrics for synthetic (generative) EHR data, covering privacy,
+utility, and statistical fidelity.
+
+.. currentmodule:: pyhealth.metrics.generative
+
+.. autofunction:: evaluate_synthetic_ehr
+
+Privacy metrics
+-------------------------------------
+
+.. autofunction:: calc_nnaar
+
+.. autofunction:: calc_membership_inference
+
+.. autofunction:: compute_discriminator_privacy
+
+Utility and fidelity metrics
+-------------------------------------
+
+.. autofunction:: compute_mle
+
+.. autofunction:: compute_prevalence_metrics
--- a/examples/halo_mimic3.py
+++ b/examples/halo_mimic3.py
@@ -1,0 +1,120 @@
+"""Example: train HALO on MIMIC-III and generate synthetic patients.
+
+This example demonstrates:
+1. Loading MIMIC-III data
+2. Applying the EHRGenerationMIMIC3 task (per-visit ICD-9 code sequences)
+3. Creating a SampleDataset with a NestedSequenceProcessor
+4. Training the HALO generator with its custom training loop
+5. Generating synthetic patients
+6. Evaluating the synthetic data with the generative metrics suite
+"""
+
+import pandas as pd
+
+from pyhealth.datasets import MIMIC3Dataset, split_by_patient
+from pyhealth.metrics.generative import evaluate_synthetic_ehr
+from pyhealth.models import HALO
+from pyhealth.tasks import EHRGenerationMIMIC3
+
+if __name__ == "__main__":
+    # STEP 1: Load MIMIC-III base dataset
+    base_dataset = MIMIC3Dataset(
+        root="/srv/local/data/MIMIC-III/mimic-iii-clinical-database-1.4",
+        tables=["diagnoses_icd"],
+        dev=True,
+    )
+
+    # STEP 2: Apply the EHR generation task (unconditional, no labels).
+    # This task is shared by all generators in pyhealth.models.generators.
+    sample_dataset = base_dataset.set_task(EHRGenerationMIMIC3())
+    print(f"Total samples: {len(sample_dataset)}")
+    print(f"Input schema: {sample_dataset.input_schema}")
+    print(f"Output schema: {sample_dataset.output_schema}")
+
+    sample = sample_dataset[0]
+    print("\nSample structure:")
+    print(f"  Patient ID: {sample['patient_id']}")
+    print(f"  Visits tensor shape: {tuple(sample['visits'].shape)}")
+
+    # STEP 3: Split dataset by patient
+    train_dataset, val_dataset, test_dataset = split_by_patient(
+        sample_dataset, [0.8, 0.1, 0.1]
+    )
+
+    # STEP 4: Initialize HALO (small config for the dev subset)
+    model = HALO(
+        dataset=sample_dataset,
+        embed_dim=128,
+        n_heads=4,
+        n_layers=4,
+        n_ctx=48,
+        batch_size=16,
+        epochs=5,
+        lr=1e-4,
+        save_dir="./halo_save",
+    )
+    num_params = sum(p.numel() for p in model.parameters())
+    print(f"\nModel initialized with {num_params} parameters")
+
+    # STEP 5: Train with HALO's custom loop (saves best checkpoint to save_dir)
+    model.train_model(train_dataset, val_dataset=val_dataset)
+
+    # STEP 6: Generate synthetic patients (one per real training patient).
+    synthetic = model.generate(num_samples=len(train_dataset), random_sampling=True)
+    print("\nGenerated synthetic patients (first 3):")
+    for patient in synthetic[:3]:
+        print(f"  {patient['patient_id']}: {len(patient['visits'])} visits")
+        print(f"    {patient['visits']}")
+
+    # STEP 7: Evaluate the synthetic data with the generative metrics suite.
+    # evaluate_synthetic_ehr expects flat dataframes with one row per code:
+    #   columns = [id, time, visit_codes, labels]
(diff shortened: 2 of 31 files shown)


pyhealth/metrics/generative/privacy.py
```diff
@@ -0,0 +1,335 @@
...
+
+__all__ = [
+    "calc_nnaar",
+    "calc_membership_inference",
+    "compute_discriminator_privacy",
+]
+
+
+def calc_nnaar(
+    train_ehr: pd.DataFrame,
```
jhnwu3, line 45: Is there a reason why we do this in dataframes? Maybe, we can chat about this.

chufangao: Yeah we could change it to the nested sequence actually since for this task, it all reduces down to a sequence of sequences. Edit, actually maybe we should keep it for utility calculation.

examples/halo_mimic3.py
```diff
@@ -0,0 +1,120 @@
...
+        f"synthetic: {len(syn_df)}"
+    )
+
+    # sample_size / n_bootstraps / n_runs are kept small for the dev subset;
+    # raise them when running on the full MIMIC-III cohort.
+    results = evaluate_synthetic_ehr(
+        train_ehr=train_df,
+        test_ehr=test_df,
+        syn_ehr=syn_df,
+        sample_size=min(30, len(train_dataset), len(test_dataset)),
```
jhnwu3, line 123: Just wondering why dataframes and why not use List[] instead?

pyhealth/metrics/generative/utility.py
```diff
@@ -0,0 +1,256 @@
...
+    "compute_mle",
+    "compute_prevalence_metrics",
+]
+
+
+def compute_mle(
+    train_fn: Callable,
+    train_ehr: pd.DataFrame,
+    test_ehr: pd.DataFrame,
+    syn_ehr: pd.DataFrame,
```
jhnwu3, line 41: I'm wondering if we can simplify to  lists of your generated samples. Is tehre a benefit to using dataframes in our evals? 

No need to do use any pyhealth primitives since we're basically at this point approaching raw data comparisons here.

jhnwu3 (changes requested):
Just wondering if there's a simpler approach here rather than using dataframes. I think it's not necessarily a problem, but it does feel like we need to add examples to our docstrings here on what are we expecting our input shapes to be, our output shapes etc.

Like we have these column args here, I think it might just be more of a documentation issue than anything.

I think a List[dict] is fine if we just have some assertions like assert visit_col is in List[0] (our sample here), so that the user knows they must define these things. The other thing is just making sure our auto-docs get the correct level of documentation here.

jhnwu3 (changes requested):
one more things can you add all your generative models to the docs.

docs/api/metrics/pyhealth.metrics.generative.rst
```diff
@@ -0,0 +1,25 @@
+pyhealth.metrics.generative
+===================================
+
+Evaluation metrics for synthetic (generative) EHR data, covering privacy,
+utility, and statistical fidelity.
+
+.. currentmodule:: pyhealth.metrics.generative
+
```
jhnwu3, line 8: It would be good to have tasks updated with your EHR generation task.

jhnwu3 (changes requested):
To update docs, simply update models.rst and then api/models/<generative>.rst. Same thing with tasks.

jhnwu3 (approved):
lgtm.

We can iterate if we find things break as things go on. Very excited to start running/testing this out.

jhnwu3:
the CI error is because of some defunct package dependency from our NLP metrics. I have a fix in: https://github.com/sunlabuiuc/PyHealth/pull/1155

PR #1150, opened 2026-05-25
Exclude pediatric patients (younger than 18) from clinical prediction tasks

DarylOkeke:
Multiple clinical prediction tasks had TODO comments saying that patients under 18 should be excluded, but the filtering was never implemented.

This adds age filtering to the mortality prediction, length-of-stay, and drug recommendation tasks across MIMIC-III, MIMIC-IV, eICU, and OMOP datasets. Adult prediction benchmarks shouldn't mix in pediatric cohorts b/c the clinical patterns are too different.

Some notes:
- eICU stores age as a string and uses "> 89" for patients over 89, so that case is handled explicitly
- MIMIC-III requires deriving age from dob and admission time since it doesn't store a precomputed age field
- All checks fail-open: if age data is missing or unparseable, the patient is included instead of dropped

### DIFF
--- a/pyhealth/tasks/drug_recommendation.py
+++ b/pyhealth/tasks/drug_recommendation.py
@@ -339,7 +339,7 @@ class DrugRecommendationMIMIC4(BaseTask):
         return samples
 
 
-def drug_recommendation_mimic3_fn(patient: Patient):
+def drug_recommendation_mimic3_fn(patient: Patient, exclude_minors: bool = True):
     """Processes a single patient for the drug recommendation task.
 
     Drug recommendation aims at recommending a set of drugs given the patient health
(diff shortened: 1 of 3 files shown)


jhnwu3 (changes requested):
Hey nice first PR! 

Is it possible to add a Boolean arg for this filtering? I think it'd be nice to be able to quickly ablation study this difference in patient demographics later.

DarylOkeke:
Thanks! I added exclude_minors to all the affected classes and legacy functions in the 3 task files. Defaults to true so the existing behavior stays the same

Another thing though, I saw that LengthOfStayPredictioneICU doesn't have an age check at all. The original TODOs didn't cover it but the other 3 LOS classes do. Want me to add one for consistency, or no?

jhnwu3:
Yeah that would be nice

DarylOkeke:
Finished!

jhnwu3:
Will check back after unit tests finish this weekend

DarylOkeke:
Hey just noticed there is a test in test_drug_recommendation_atc3.py that checks drug list order with assertEqual but the list comes back in a different order on CI should I swap it to assertCountEqual so it's order-independent, or leave it?

jhnwu3 (changes requested):
One last thing before we meet, can you verify it works locally on your machine with our demo dataset?

https://physionet.org/content/mimic-iv-demo/2.2/

jhnwu3:
closed bc duplicate (it seems?). Reopen if you disagree

PR #1151, opened 2026-05-26
Fix trainer mode resolution (closes #914)

### LINKED ISSUE #914: PyHealth trainer uses deprecated model.mode attribute instead of checking its output or label processors
https://github.com/sunlabuiuc/PyHealth/blob/8d71772bbfb01e5fb8ec26528f07139d00f02a86/pyhealth/trainer.py#L325 

Basically, we set a mode for what metrics we use, binary, multiclass, multilabel, etc. 

The irony here is we can simply just directly map that mode with our output processors and thereby eliminating an intermediate variable.

This isn't a big deal now, but would be good to remedy as we start incorporating things like reconstruction losses, generative modeling techniques, etc. where we want to scale what metrics are compatible with not only the vanilla PyHealth trainer but also things like PyTorch lightning that is directly compatible with the litData SampleDataset generated by any set_task()

DarylOkeke:
Replaces model.mode with _resolve_mode(output_schema) in Trainer.evaluate, falling back to model.mode for custom models.

### DIFF
--- a/pyhealth/tasks/drug_recommendation.py
+++ b/pyhealth/tasks/drug_recommendation.py
@@ -339,7 +339,7 @@ class DrugRecommendationMIMIC4(BaseTask):
         return samples
 
 
-def drug_recommendation_mimic3_fn(patient: Patient):
+def drug_recommendation_mimic3_fn(patient: Patient, exclude_minors: bool = True):
     """Processes a single patient for the drug recommendation task.
 
     Drug recommendation aims at recommending a set of drugs given the patient health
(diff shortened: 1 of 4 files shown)


jhnwu3 (commented):
These functions should be deleted at some point. They're redundant and don't actually work in the framework lol.

jhnwu3:
closing b/c we'll be deleting these old legacy functions soon (tbh we completely forgot lol).

jhnwu3:
ah nvm I see this is your newer PR. Let's close the other one.

PR #1177, opened 2026-07-08
Add EEGBCI dataset and tasks

vihaan101:
**Contributor:** Vihaan Agrawal

**Contribution Type:** Dataset + Task + Example + Documentation

**Description:**
Adds support for the PhysioNet EEG Motor Movement/Imagery (EEGBCI) dataset as a PyHealth EEG pipeline.

- **Dataset:** Adds `EEGBCIDataset`, dataset configuration, local EDF discovery, optional MNE-based download, and metadata generation for selected subjects and runs.
- **Task:** Adds EEGBCI motor execution and motor imagery label mapping, channel selection, normalization, band-power features, and pattern discovery sample generation.
- **Example:** Adds an EEGBCI pattern discovery example and README for running the pipeline.
- **Documentation:** Registers the new dataset and task in the API docs.
- **Tests:** Adds core coverage for dataset metadata preparation, download handling, label mapping, signal helpers, feature extraction, and task integration.

**Data Access:**
EEGBCI files are loaded from local EDF files when available, or fetched through MNE with `download=True`.

**Files to Review:**

| File | Description |
|---|---|
| `pyhealth/datasets/eegbci.py` | EEGBCI dataset loader and metadata preparation |
| `pyhealth/datasets/configs/eegbci.yaml` | Dataset table configuration |
| `pyhealth/tasks/eegbci.py` | EEGBCI task labels, signal helpers, feature extraction, and pattern discovery task |
| `examples/eeg/eegbci/eegbci_pattern_discovery.py` | End-to-end pattern discovery example |
| `examples/eeg/eegbci/README.md` | Example usage notes |
| `docs/api/datasets/pyhealth.datasets.EEGBCIDataset.rst` | Dataset API documentation |
| `docs/api/tasks/pyhealth.tasks.eegbci.rst` | Task API documentation |
| `tests/core/test_eegbci.py` | Dataset and task test coverage |

**Testing:**

- `make testall`
  - core: 1179 tests passed, 73 skipped
  - NLP: 2 tests passed


pyhealth/datasets/eegbci.py
```diff
@@ -0,0 +1,139 @@
...
+        runs: Optional[list[int]] = None,
+        download: bool = False,
+        **kwargs,
+    ) -> None:
+        if config_path is None:
+            config_path = Path(__file__).parent / "configs" / "eegbci.yaml"
+        self.root = root
+        self.subjects = list(subjects) if subjects is not None else [1, 2, 3]
+        self.runs = list(runs) if runs is not None else list(range(3, 15))
+        self.download = download
```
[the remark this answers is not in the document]
vihaan101: Addressed in 712619a. `EEGBCIDataset` now normalizes `subjects` and `runs` to sorted unique integers before building `selection_key`, `dataset_name`, or metadata, with focused regression coverage for duplicate/reordered selections.

pyhealth/datasets/eegbci.py
```diff
@@ -0,0 +1,139 @@
...
+        ).hexdigest()[:10]
+        subject_part = "-".join(f"{int(subject):03d}" for subject in self.subjects)
+        run_part = "-".join(f"{int(run):02d}" for run in self.runs)
+        return f"s{subject_part}_r{run_part}_{digest}"
+
+    def _find_local_edf(self, subject: int, run: int) -> Path | None:
+        root = Path(self.root)
+        pattern = f"S{subject:03d}R{run:02d}.edf"
+        matches = sorted(root.rglob(pattern))
+        return matches[0] if matches else None
```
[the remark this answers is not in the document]
vihaan101: Addressed in 712619a. `_find_local_edf()` now checks the canonical MNE EEGBCI path first and only falls back to recursive lookup for nonstandard cache layouts.

pyhealth/datasets/eegbci.py
```diff
@@ -0,0 +1,139 @@
...
+        if not EEGBCI_METADATA_COLUMNS.issubset(df.columns):
+            return False
+        pairs = sorted((int(row.subject_id), int(row.run)) for row in df.itertuples())
+        return pairs == self._requested_pairs()
+
+    def prepare_metadata(self) -> None:
+        root = Path(self.root)
+        csv_path = root / "eegbci-pyhealth.csv"
+        if csv_path.exists() and self._metadata_matches_request(csv_path):
+            return
```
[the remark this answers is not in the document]
vihaan101: Addressed in 712619a. Metadata is now written to a selection-specific CSV filename derived from `selection_key`, and the loaded dataset config points at that file so separate selections do not overwrite each other.

examples/eeg/eegbci/eegbci_pattern_discovery.py
```diff
@@ -0,0 +1,649 @@
...
+
+def parse_int_list(value: str) -> list[int]:
+    items: list[int] = []
+    for part in value.split(","):
+        if "-" in part:
+            start, end = part.split("-", 1)
+            items.extend(range(int(start), int(end) + 1))
+        else:
+            items.append(int(part))
+    return items
```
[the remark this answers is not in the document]
vihaan101: Addressed in 712619a. `parse_int_list()` now strips each comma-separated part, rejects empty entries, and raises a clear `ValueError` for descending ranges like `5-3`.

examples/eeg/eegbci/eegbci_pattern_discovery.py
```diff
@@ -0,0 +1,649 @@
...
+    summary_path = output_dir / "eegbci_pattern_summary.md"
+    pd.DataFrame(output_rows, columns=OUTPUT_COLUMNS).to_csv(csv_path, index=False)
+    write_summary(
+        output_rows,
+        summary_path,
+        {
+            "subjects": parse_int_list(args.subjects),
+            "runs": parse_int_list(args.runs),
+            "max_windows": args.max_windows,
+            "baseline_row_count": len(all_rows),
```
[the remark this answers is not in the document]
vihaan101: Addressed in 712619a. The report now computes `baseline_row_count` from rows with `task_label == "rest"`, so “Baseline source rows” reflects the rest windows used for baseline construction instead of all windows.

pyhealth/tasks/eegbci.py
```diff
@@ -0,0 +1,373 @@
...
+        "brain_state_hypothesis": hypothesis,
+        "confidence": confidence,
+        "quality_flags": ";".join(quality_flags) if quality_flags else "none",
+        "interpretation": (
+            f"The segment is consistent with {hypothesis} based on a "
+            f"{dominant}-dominant frequency profile."
+        ),
+    }
+
+
```
Rian354, line 382: The interpretation labels come out constant on the demo run.

 Running the documented command (--subjects 1 --runs 3), all 20 windows land on slow_wave_dominant_pattern/delta, same label for rest, execute_left_fist, and execute_right_fist. In the code, dominant_band is argmax of absolute band power (line 185), which is delta on real EEG, and the alpha_relative >= 0.45 / beta_relative >= 0.35 branches (lines 208-217) are only hit by the synthetic-sinusoid tests and not the real signals, so the state output doesn't vary by condition.
  
The example also averages power across all channels (line 169), which drops the C3/C4 information. Could we drive the state off the rest normalized deltas the example already computes so the output varies and the branches are reachable?

pyhealth/tasks/eegbci.py
```diff
@@ -0,0 +1,373 @@
...
+    "rest": 0,
+    "execute_left_fist": 1,
+    "execute_right_fist": 2,
+    "imagine_left_fist": 3,
+    "imagine_right_fist": 4,
+    "execute_both_fists": 5,
+    "execute_both_feet": 6,
+    "imagine_both_fists": 7,
+    "imagine_both_feet": 8,
+}
```
Rian354, line 38: Since label is multiclass, it goes through MultiClassLabelProcessor, which only preserves the 0-8 map when the labels are contiguous. left_right runs ({0,1,2}) are fine, but fists/feet runs ({0,5,6}, {0,7,8}) get remapped, so execute_both_fists becomes trained index 1, the same index execute_left_fist gets in a different selection. Giving task_label (string) to the processor would probably handle this

pyhealth/datasets/eegbci.py
```diff
@@ -0,0 +1,158 @@
...
+        df = pd.DataFrame(rows)
+        df.sort_values(["subject_id", "run"], inplace=True)
+        df.reset_index(drop=True, inplace=True)
+        csv_path.parent.mkdir(parents=True, exist_ok=True)
+        df.to_csv(csv_path, index=False)
+        logger.info("Wrote EEGBCI metadata to %s", csv_path)
+
+    @property
+    def default_task(self) -> EEGBCIPatternDiscovery:
+        return EEGBCIPatternDiscovery()
```
Rian354: Other datasets return the canonical supervised task, so a bare set_task() here does the extra band power/interpretation work and emits fields that no processor uses. Would returning EEGMotorImageryEEGBCI() (keeping EEGBCIPatternDiscovery for the example) work here?

pyhealth/datasets/eegbci.py
```diff
@@ -0,0 +1,158 @@
...
+        if not EEGBCI_METADATA_COLUMNS.issubset(df.columns):
+            return False
+        pairs = sorted((int(row.subject_id), int(row.run)) for row in df.itertuples())
+        return pairs == self._requested_pairs()
+
+    def prepare_metadata(self) -> None:
+        root = Path(self.root)
+        csv_path = root / self.metadata_file_name
+        if csv_path.exists() and self._metadata_matches_request(csv_path):
+            return
```
Rian354, line 202: _metadata_matches_request checks columns and (subject, run) pairs but not that the signal_file paths still exist, so a stale CSV is reused and fails later in read_raw_edf instead of at the loader's FileNotFoundError. A Path(row.signal_file).exists() check would catch it in advance.

pyhealth/datasets/eegbci.py
```diff
@@ -0,0 +1,158 @@
...
+    "record_id",
+    "subject_id",
+    "run",
+    "run_type",
+    "signal_file",
+    "source",
+}
+
+
+class EEGBCIDataset(BaseDataset):
```
Rian354, line 28: The contributing guide asks for Google style docstrings. The task classes, dataset class, and helpers don't have them yet, and the .rst uses :undoc-members: so they render bare. Going over the constructor args (channel_mode, normalization, compute_stft, resample_rate/bandpass_filter) and emitted keys would cover it

pyhealth/tasks/eegbci.py
```diff
@@ -0,0 +1,373 @@
+from __future__ import annotations
```
Rian354, line 3: Docstring reminder, see datasets/eegbci.py line 28.

tests/core/test_eegbci.py
```diff
@@ -0,0 +1,1522 @@
...
+                subjects=[1],
+                runs=[3],
+                download=False,
+                cache_dir=root / "cache",
+            )
+
+            with patch("pyhealth.tasks.eegbci.mne.io.read_raw_edf", return_value=raw):
+                sample_dataset = dataset.set_task(
+                    EEGMotorImageryEEGBCI(
+                        compute_stft=False, resample_rate=None, bandpass_filter=None
```
Rian354: The task tests all pass resample_rate=None, bandpass_filter=None, so raw.filter()/raw.resample() (the defaults) never run, a case with defaults on, asserting window length, would cover them. The STFT test mocks the core computation, so it doesn't validate the implementation itself; an unmocked assertion on sample["stft"]'s shape would help.

examples/eeg/eegbci/README.md
```diff
@@ -0,0 +1,56 @@
...
+fields. Legacy task-level fields such as `brain_state_hypothesis`, `confidence`,
+and `quality_flags` are intentionally not written to the CSV.
+
+The Markdown report summarizes state counts, task-label/state agreement,
+rest-normalized bandpower deltas, confidence and quality flags, representative
+windows, limitations, and next checks. These labels are signal-pattern
+summaries from short EEG windows, not clinical findings or evidence of a
+subject's cognition.
+
+Implementation details are tracked in
```
Rian354: The "Implementation details are tracked in …" line links a file that isn't in the PR, probably fine to drop. 

Could you also add the PhysioNet source (eegmmidb), license (ODC-BY 1.0), and citation to the README or dataset docstring?

Rian354 (commented):
Thanks for your PR. The unit tests pass, the real data smoke test downloads S001R03 fine, and the example runs e2e. The dataset loader, caching, and channel handling look solid. A few things need to be addressed before merge.

Rian354:
Before merge, can you confirm it runs on your end with the demo command as well?

pyhealth/tasks/eegbci.py
vihaan101: Thanks for reviewing this and also checking the documented run. Changed this to use the rest-normalized relative-band deltas when they are available, with the absolute band profile only as fallback. The documented S001/R03 run now gives 4 states across the 20 windows instead of all slow-wave: 11 slow-wave, 5 sensorimotor, 2 ambiguous, and 2 idle. Also added a finite-value check so NaN/inf deltas use the fallback.

pyhealth/tasks/eegbci.py
vihaan101: Changed `label` to use the semantic `task_label` string for `MultiClassLabelProcessor`. The stable 0-8 PyHealth identifier is still kept separately as `eegbci_label`, so sparse fists/feet selections no longer overlap with labels from another selection.

pyhealth/datasets/eegbci.py
vihaan101: Yes, changed `default_task` to return `EEGMotorImageryEEGBCI()`. `EEGBCIPatternDiscovery` is still available explicitly and the example continues using it, so bare `set_task()` now only does the canonical supervised task work.

pyhealth/datasets/eegbci.py
vihaan101: Added the signal path check before reusing the metadata CSV. If a referenced EDF is gone, the metadata is rebuilt when another matching EDF exists, otherwise it reaches the dataset `FileNotFoundError` instead of failing later inside `read_raw_edf`. The metadata contents are also part of the derived cache identity now.

pyhealth/datasets/eegbci.py
vihaan101: Added Google style docstrings to the dataset class and its helpers. The constructor arguments, metadata behavior, source, license, and possible `FileNotFoundError` are covered now.

pyhealth/tasks/eegbci.py
vihaan101: Added the module, task class, and public helper docstrings here too. The preprocessing arguments and emitted fields including `task_label`, `label`, `eegbci_label`, `signal`, and optional `stft` are documented.

tests/core/test_eegbci.py
vihaan101: Added an integration case using the default task path, so filtering and 160 to 200 Hz resampling run and the two-second window is checked as 400 samples. The STFT assertion now uses the real computation instead of mocking it and checks the `(16, 50, 3)` output shape.

examples/eeg/eegbci/README.md
vihaan101: Dropped the missing implementation-plan link. Added the PhysioNet eegmmidb source, ODC-By 1.0 license, and the Schalk, BCI2000, and standard PhysioNet citations. The dataset docstring also includes the source, license, and DOI.

vihaan101:
@Rian354 Yes, confirmed on my end with the documented command:

```bash
python examples/eeg/eegbci/eegbci_pattern_discovery.py \
  --subjects 1 \
  --runs 3 \
  --max-windows 20 \
  --download
```

It completed and wrote both the CSV and Markdown summary. I got 20 windows with 4 states: 11 slow-wave, 5 sensorimotor, 2 ambiguous, and 2 idle. All 20 rows also have an evidence score.

Rian354 (commented):
LGTM. Thanks for you hard work, appreciate the fast commit on the contrib rules as well.

Rian354 (approved):

Issue #329, opened 2025-04-09
Integration with MEDS


zzachw:
Integrate with the MEDS ecosystem (https://medical-event-data-standard.github.io/docs/intro)

PR #1179, opened 2026-07-15
Add MEDS dataset support (MEDSDataset + typed Parquet scan path)

AxelNoun:
## Motivation

The [Medical Event Data Standard (MEDS)](https://github.com/Medical-Event-Data-Standard/meds) is a minimal, event-based schema for machine learning over EHR data (MEDS Working Group, *Medical Event Data Standard (MEDS): Facilitating Machine Learning for Health*, ICLR 2024 Workshop on Learning from Time Series For Health, [openreview:IsHy2ebjIG](https://openreview.net/forum?id=IsHy2ebjIG)). MEDS datasets are distributed as sharded, natively typed Parquet files.

PyHealth currently cannot read Parquet sources: `.parquet` paths and directories fall through the CSV/TSV extension resolver and raise `ValueError`. Since the MEDS core table maps one-to-one onto PyHealth's canonical event frame (`subject_id` → `patient_id`, `time` → `timestamp`, `code`/`numeric_value` as attributes), MEDS support is almost purely declarative once Parquet can be scanned.

Related to #329 (Integration with MEDS).

## Changes

**1. `BaseDataset`: generic Parquet source support (~90 changed lines)**

- `_scan_table`: format router — `.parquet`/`.pq` files, glob patterns, and directories go to the new scanner; everything else falls back to the existing `_scan_csv_tsv_gz`, unchanged. Non-breaking by construction: all newly routed inputs previously raised.
- `_scan_parquet`: typed scan (no all-string coercion — Parquet embeds its schema) with the same Dask kwargs as the cached-Parquet read of the CSV path. A directory is scanned recursively, which supports the sharded MEDS layout `data/<split>/<shard>.parquet`. Missing or Parquet-less sources raise `FileNotFoundError`.
- `load_table`: a fast-path when the configured timestamp column is already datetime-typed — skips the string round-trip, preserves `NaT` (MEDS static events have null `time`), then applies the usual `datetime64[ms]` cast. The list-concatenation and string-parsing branches are untouched.
- A `>>>` usage example was added to the `BaseDataset` docstring, as required by the PR rules once the class span is modified.

**2. `MEDSDataset` + `configs/meds.yaml`**

- Declarative wrapper patterned on `EHRShotDataset`; an optional static `subject_splits` table mirrors the EHRShot `splits` precedent.
- `subset="train" | "tuning" | "held_out"`, resolved via `metadata/subject_splits.parquet` (with a directory-layout fallback); each subset gets its own processing cache. Linked in source docs and the API reference page to the [MEDS schema documentation](https://medical-event-data-standard.github.io/).
- A construction-time schema guard reads Parquet footers only (no data, no Dask): a missing, non-timestamp, or timezone-aware `time` column raises `TypeError`. The MEDS reference schema is timezone-naive `timestamp[us]`; the guard also rejects date-like integer columns (e.g. `20240101`) that the generic string fallback would otherwise parse silently.

**3. Tests — `tests/core/test_meds.py`**

- Deterministic synthetic sharded fixtures: nested split directories, static null-time events, µs→ms precision, subset filtering, distinct caches per subset, schema-guard violations. CI needs no external data.
- Demo-backed smoke tests are skip-gated behind `MEDS_DEMO_ROOT` (public *MIMIC-IV demo data in MEDS*, PhysioNet, ODbL v1.0).
- The CSV path shared by existing datasets is regression-covered by the current suite.

**4. Docs & example**

- `docs/api/datasets/pyhealth.datasets.MEDSDataset.rst`
- `examples/meds_demo.py`: load/split peek, then end-to-end `InHospitalMortalityMEDS` → `RNN` → `Trainer` on the public demo (238 stays, 80/10/10 split). Full-stay code sequences are tail-truncated to 256 events in a demo-only wrapper so the script finishes on CPU (~2.5 min, mostly `set_task` I/O); not a benchmark.

**5. `InHospitalMortalityMEDS` (MEDS-native task)**

- One sample per completed stay: group events on `hadm_id` from `HOSPITAL_ADMISSION//*` / `HOSPITAL_DISCHARGE//*`, label = `HOSPITAL_DISCHARGE//DIED`, features = ordered codes in half-open `[admit, prediction_time)` excluding discharge and `MEDS_DEATH` (anti-leakage). Default window `full_stay`; opt-in `first_hours`.
- `hadm_id` is not core MEDS, so it stays out of `configs/meds.yaml` and is exposed via bundled `configs/meds_with_hadm.yaml` (absent attributes raise in `load_table`).
- Tests: synthetic leakage/window/label coverage + optional demo smoke; `examples/verify_meds_mortality.py` prints cohort stats through the real pipeline.
- API docs: `docs/api/tasks/pyhealth.tasks.InHospitalMortalityMEDS.rst` + toctree entry in `docs/api/tasks.rst`.

## Testing

- `python -m pytest tests/core/test_meds.py tests/core/test_in_hospital_mortality_meds.py -v`: **24 passed** (incl. schema-guard subtests + mortality synthetic/demo smoke) — Windows, Python 3.12.
- Local smoke on the public demo: 100 subjects / 803,992 events; canonical dtypes verified (`patient_id` string, `timestamp` `datetime64[ms]`).
- **Cohort via real task pipeline** (`examples/verify_meds_mortality.py --root <demo>`): **12 positive / 238 stays** (rate 0.0504); `set_task` sample count **238**.
- **`examples/meds_demo.py`** (optional, requires demo download): completes on CPU in ~2.5 min; training epoch ~12 s on truncated sequences.

## How to verify

```bash
pip install -e ".[dev]"
python -m pytest tests/core/test_meds.py tests/core/test_in_hospital_mortality_meds.py -v

# Optional: demo-backed tests + examples on the public PhysioNet demo (ODbL)
wget -r -N -c -np https://physionet.org/files/mimic-iv-demo-meds/0.0.1/
MEDS_DEMO_ROOT=physionet.org/files/mimic-iv-demo-meds/0.0.1 \
    python -m pytest tests/core/test_meds.py tests/core/test_in_hospital_mortality_meds.py -v
python examples/meds_demo.py --root physionet.org/files/mimic-iv-demo-meds/0.0.1
python examples/verify_meds_mortality.py --root physionet.org/files/mimic-iv-demo-meds/0.0.1
```

## Notes for review

- **Scope**: the `BaseDataset` Parquet path is deliberately generic (reusable by future Parquet-distributed datasets). Happy to split it into its own PR if you prefer smaller diffs.
- **`subset` API**: small and tested, but easy to strip if you want the first cut minimal.
- **Demo-backed tests**: currently skip-gated so the repo carries no new data. Alternatively I can commit a reduced ODbL slice (+ LICENSE/NOTICE) under `test-resources/`, following the eICU/MIMIC demo precedent — your call.
- **`hadm_id` config**: stay-aware tasks opt into `meds_with_hadm.yaml`; the default MEDS config stays core-schema-only so generic MEDS exports without `hadm_id` keep working.

### DIFF
--- a/examples/meds_demo.py
+++ b/examples/meds_demo.py
@@ -1,0 +1,76 @@
+"""End-to-end example: loading a MEDS dataset with PyHealth.
+
+This example uses the public *MIMIC-IV demo data in the Medical Event Data
+Standard (MEDS)* (PhysioNet, v0.0.1, ODbL v1.0, ~100 subjects):
+https://doi.org/10.13026/t2y8-ea41
+
+Download it once (open access, ~a few MB):
+
+    wget -r -N -c -np https://physionet.org/files/mimic-iv-demo-meds/0.0.1/
+
+Then run:
+
+    python examples/meds_demo.py \\
+        --root physionet.org/files/mimic-iv-demo-meds/0.0.1
+
+Any dataset following the MEDS layout (``data/**.parquet`` +
+``metadata/subject_splits.parquet``) works the same way. See the MEDS
+specification: https://github.com/Medical-Event-Data-Standard/meds
+"""
+
+import argparse
+
+import polars as pl
+
+from pyhealth.datasets import MEDSDataset
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--root",
+        required=True,
+        help="Root of the MEDS dataset (directory containing data/ and "
+        "metadata/)",
+    )
+    parser.add_argument(
+        "--subset",
+        default="train",
+        help="Split to load as a subset (default: train)",
+    )
+    args = parser.parse_args()
+
+    # 1) Load the full dataset: every Parquet shard under data/ is read,
+    #    including nested split directories (data/<split>/<shard>.parquet).
+    dataset = MEDSDataset(root=args.root)
+    dataset.stats()
+
+    # 2) Peek at the canonical event frame (typed straight from Parquet:
+    #    string patient ids, datetime64[ms] timestamps, float values).
+    events = dataset.global_event_df
+    print(events.head(5).collect())
+
+    # 3) Load a split-restricted subset. Subjects are selected through the
+    #    metadata/subject_splits.parquet assignment; each subset uses its
+    #    own processing cache.
+    subset = MEDSDataset(root=args.root, subset=args.subset)
+    n_subset = len(subset.unique_patient_ids)
+    n_total = len(dataset.unique_patient_ids)
+    print(f"Subjects in subset '{args.subset}': {n_subset} / {n_total}")
+
+    # 4) Static (null-time) MEDS events, e.g. demographics, are preserved.
+    n_static = (
+        events.filter(pl.col("timestamp").is_null())
+        .select(pl.len())
+        .collect()
+        .item()
+    )
+    print(f"Static (null-time) events: {n_static}")
+
+    # From here, the dataset behaves like any other PyHealth dataset: use
+    # `dataset.set_task(...)` with an existing task to build samples.
+
+
+if __name__ == "__main__":
+    # BaseDataset spawns Dask worker processes; keep the main-module guard.
+    main()
--- a/examples/meds_demo.py
+++ b/examples/meds_demo.py
@@ -125,7 +125,12 @@ def main() -> None:
         epochs=1,
         monitor="roc_auc",
     )
-    print(trainer.evaluate(test_dataloader))
+    metrics = trainer.evaluate(test_dataloader)
+    print(
+        "Test metrics (smoke test only — 1 epoch, tail-truncated sequences; "
+        "not interpretable as model quality):"
+    )
+    print(metrics)
 
 
 if __name__ == "__main__":
--- a/examples/meds_demo.py
+++ b/examples/meds_demo.py
@@ -19,10 +19,41 @@ specification: https://github.com/Medical-Event-Data-Standard/meds
 """
 
 import argparse
+import tempfile
+from pathlib import Path
+from typing import Any, List
 
 import polars as pl
+import pyhealth.datasets.configs as meds_configs
 
-from pyhealth.datasets import MEDSDataset
+from pyhealth.datasets import MEDSDataset, get_dataloader, split_by_patient
+from pyhealth.models import RNN
+from pyhealth.tasks import InHospitalMortalityMEDS
+from pyhealth.trainer import Trainer
+
+# Full-stay MEDS code sequences are often thousands of events long. Keeping
+# every code makes a vanilla RNN prohibitively slow on CPU, so the training
+# block below keeps only the *most recent* codes per stay. This is a demo
+# ergonomics choice, not a benchmark configuration: production runs should
+# use the unmodified task (``InHospitalMortalityMEDS()``) and an appropriate
+# model/window for the sequence lengths involved.
+_DEMO_MAX_SEQ_LEN = 256
+
+
+class _DemoMortalityTask(InHospitalMortalityMEDS):
+    """Demo-only wrapper that tail-truncates ``codes`` before ``set_task``."""
+
+    def __init__(self, max_seq_len: int = _DEMO_MAX_SEQ_LEN, **kwargs) -> None:
+        super().__init__(**kwargs)
+        self.max_seq_len = max_seq_len
+
+    def __call__(self, patient: Any) -> List[dict]:
+        samples = super().__call__(patient)
+        for sample in samples:
+            codes = sample["codes"]
+            if len(codes) > self.max_seq_len:
+                sample["codes"] = codes[-self.max_seq_len :]
+        return samples
 
 
 def main() -> None:
@@ -30,8 +61,7 @@ def main() -> None:
     parser.add_argument(
         "--root",
         required=True,
-        help="Root of the MEDS dataset (directory containing data/ and "
-        "metadata/)",
+        help="Root of the MEDS dataset (directory containing data/ and metadata/)",
     )
     parser.add_argument(
         "--subset",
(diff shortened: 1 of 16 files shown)


pyhealth/datasets/meds.py
```diff
@@ -0,0 +1,301 @@
+"""MEDS (Medical Event Data Standard) dataset for PyHealth.
+
+MEDS distributes event data as *typed*, sharded Parquet, already flattened to
+one row per measurement -- ``(subject_id, time, code, numeric_value, ...)`` --
+plus a canonical subject-to-split mapping at
+``metadata/subject_splits.parquet``. This maps almost one-to-one onto
```
jhnwu3: Is it possible to reference this split mapping here to a MEDS doc link? Just asking to make sure users who read the docs know about this.

pyhealth/tasks/in_hospital_mortality_meds.py
```diff
@@ -0,0 +1,287 @@
...
+                integer ``_hadm`` column already attached.
+
+        Returns:
+            pl.DataFrame: Columns ``_hadm``, ``admit``, ``discharge``,
+            ``discharge_code``, one row per ``hadm_id`` that has both an
+            admission and a discharge. Malformed duplicates collapse via
+            earliest-admission / latest-discharge aggregation.
+        """
+        code = pl.col("meds/code")
+        admissions = (
```
jhnwu3, line 176: ah, I see what you're doing here. Was going to ask for more clarification, but maybe "aggregate_stays" might be a better term. Reconstruction implies something like actual reconstruction loss, etc. here.

pyhealth/tasks/in_hospital_mortality_meds.py
```diff
@@ -0,0 +1,287 @@
...
+                    "mortality": int(stay["discharge_code"] == DIED_CODE),
+                }
+            )
+
+        return samples
+
+    @staticmethod
+    def summarize(samples: List[Dict[str, Any]]) -> Dict[str, Any]:
+        """Summary statistics of a generated sample set, for cohort reporting.
+
```
jhnwu3: I'm curious what this is used for? For reference, the BaseDataset.set_task() -> SampleDataset which is a PyTorch data class, meaning it's not exactly a List[Dict] here.

examples/meds_demo.py
```diff
@@ -0,0 +1,76 @@
...
+    print(events.head(5).collect())
+
+    # 3) Load a split-restricted subset. Subjects are selected through the
+    #    metadata/subject_splits.parquet assignment; each subset uses its
+    #    own processing cache.
+    subset = MEDSDataset(root=args.root, subset=args.subset)
+    n_subset = len(subset.unique_patient_ids)
+    n_total = len(dataset.unique_patient_ids)
+    print(f"Subjects in subset '{args.subset}': {n_subset} / {n_total}")
+
```
jhnwu3, line 90: Can we try training an RNN model at all on this? from pyhealth.models. That would help us showcase it works.

jhnwu3 (changes requested):
Sorry for the late reply! I've got a couple small nitpicks, but yeah this is basically how I would've done it!

I've got 4 small comments, thanks for the hard work!

pyhealth/datasets/meds.py
AxelNoun: Added: linked the MEDS schema docs next to the subject_splits.parquet mention.

pyhealth/tasks/in_hospital_mortality_meds.py
AxelNoun: Agreed, reconstruction was misleading. Renamed to _group_stays; stays are grouped on hadm_id, not reconstructed. Prose updated to match.

pyhealth/tasks/in_hospital_mortality_meds.py
AxelNoun: It takes the raw List[Dict] from __call__ (per-patient application), not the tokenized SampleDataset from set_task(). Callers (examples/verify_meds_mortality.py, tests) use it that way deliberately, so cohort stats stay inspectable on the string codes before processors are fitted. Clarified the docstring. Happy to move it out of the task class if you'd rather it live elsewhere.

examples/meds_demo.py
AxelNoun: Added: set_task(InHospitalMortalityMEDS()) → pyhealth.models.RNN → Trainer, on the full demo cohort (238 stays, 80/10/10). Full-stay sequences run ~3k codes, so the demo truncates to the last 256 per stay via a local wrapper task API unchanged. ~2.5 min on CPU, mostly dataset loading rather than training. It's a smoke test that the MEDS path works with the existing stack, not a benchmark. Happy to expose the cap as a task parameter if that'd be generally useful.

AxelNoun:
Thanks for the review / all four addressed and pushed.

pyhealth/datasets/meds.py
AxelNoun: Follow-up: also added the link to the rendered API page intro (docs/api/datasets/pyhealth.datasets.MEDSDataset.rst), in addition to the class docstring in meds.py.

examples/meds_demo.py
AxelNoun: Follow-up with a local run on the PhysioNet demo (CPU, ~2.5 min total):

```
Mortality task samples (train, codes tail-truncated to 256): 238
Epoch 0 / 1: 100%|...| 7/7 [~27s]
Test metrics (smoke test only — 1 epoch, tail-truncated sequences; not interpretable as model quality):
{'pr_auc': 0.13, 'roc_auc': 0.38, 'f1': 0.22, 'loss': 0.67}
```

(Metrics are noise on this tiny demo; included only to show the path runs end-to-end.)

pyhealth/tasks/in_hospital_mortality_meds.py
jhnwu3: Oh so literally .summarize() before set_task()?

Hmm, I think you might need to change how it works, because it'll be really slow for much larger datasets as it doesn't reuse the set_task() operations and it won't be memory efficient since it's not using our lazyloading?. I think it's easier to do it post-tokenization here, because then you can leverage the parallel threading to do the extraction first and then everything is already a tensor.

jhnwu3 (changes requested):
1 last request, you can probably delete it or write a separate utiltiy function that does it for the SampleDataset instead.

pyhealth/tasks/in_hospital_mortality_meds.py
AxelNoun: Fair point;  the per-patient loop bypasses lazy loading and the parallel processing in `set_task()`, so it wouldn't scale past demo-sized data. Removed it, along with its call sites in the test and example. Happy to add a `SampleDataset`-based utility as a follow-up if cohort stats turn out to be worth having;  it probably belongs outside the task class either way.

AxelNoun:
> 1 last request, you can probably delete it or write a separate utiltiy function that does it for the SampleDataset instead.

I preferred to delete it seemed more coherent to see to add later maybe?

jhnwu3 (approved):
lgtm. Will merge when workflows finished.

AxelNoun:
@jhnwu3 I modified so that the contribution rules pass, it should be good now. Thank you for your patience !

PR #1191, opened 2026-08-19
Fix GAMENet's Dynamic Memory

lehendo:
GAMENet.forward() hardcoded prev_drugs = torch.zeros(...) instead of the patient's actual drug history

require drugs_hist in the dataset schema, precompute a remap from drugs_hist's own input vocabulary to the drugs label_vocab used by ehr_adj/ddi_adj/the Memory Bank, and build the real multi-hot prev_drugs tensor from that remapped history

### DIFF
--- a/docs/api/models/pyhealth.models.GAMENet.rst
+++ b/docs/api/models/pyhealth.models.GAMENet.rst
@@ -3,6 +3,13 @@
 
 The separate callable GAMENetLayer and the complete GAMENet model.
 
+GAMENet requires ``drugs_hist`` (nested per-visit drug history, with the
+current/target visit already zeroed out, e.g. as produced by
+:mod:`pyhealth.tasks.drug_recommendation`) in the dataset's ``input_schema``.
+This is used to populate the paper's Dynamic Memory (Eq. 6): each previous
+visit's actual administered drugs, retrieved via the query-key temporal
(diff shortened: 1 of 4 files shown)


fbonc (approved):

fbonc (changes requested):
This problem predates your changes in this PR, but i think this implementation is incorrect for mixed-length batches now that dynamic memory is actually considered.

e.g., if patient A has 4 visits and patient B has 2, padding produces:
A: A1 A2 A3 A4
B: B1 B2 PAD PAD

queries[:, :-1] removes only the final column, producing:

A: A1 A2 A3
B: B1 B2 PAD

So this (correctly) excludes A’s current visit (A4), but it leaves B’s current visit (B2) AND padding in dynamic memory. This means softmax can assign attention to these invalid positions, taking weight away from B’s actual history (B1 here), which is bad.

I think to fix use the visit mask to exclude padding and each patient’s last valid visit before softmax. Also, patients without any prior visits should receive a zero dynamic-memory result.

DarylOkeke:
The mixed-length mask looks good. One remaining issue: when a patient has no previous visits, the authors’ implementation reuses the graph-memory result, but this code uses zero and the new test locks that in. Could we match the reference behavior for first visits?

fbonc:
looks good now.

PR #1193, opened 2026-08-20
Fix patient merging on null patient_id tables (per-partition reset_index)

fbonc:
__Problem__
For tables configured with `patient_id: null` (one patient per row), the code assigned `patient_id` via `df.reset_index(drop=True)` then `df.index`. In Dask `reset_index` resets per partition, so rows at the same position in different partitions collide onto the same id and get silently merged into one "patient". This corrupts several shipped configs (`covid19_cxr`, `medical_transcriptions`, `clinvar`, eICU hospital) and defeats `split_by_patient` / `split_by_patient_conformal`, whose entire purpose is to prevent cross-patient leakage.

__Fix__
Assign a globally-unique id with a `cumsum` (global across partitions) instead of the per-partition index, in both `base_dataset.py` and `fhir/base.py`.

__Tests__
Added a regression test for each path that builds a null-`patient_id` table across multiple Dask partitions and asserts every `patient_id` is unique.

### DIFF
--- a/pyhealth/datasets/base_dataset.py
+++ b/pyhealth/datasets/base_dataset.py
@@ -749,7 +749,13 @@ class BaseDataset(ABC):
             df: dd.DataFrame = df.assign(patient_id=df[patient_id_col].astype("string"))
         else:
             df: dd.DataFrame = df.reset_index(drop=True)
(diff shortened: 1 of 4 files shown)


lehendo (approved):
Looks good to me!

PR #1194, opened 2026-08-20
Fix GRASP's dead similar-patient graph

fbonc:
__Problem__
`GRASPLayer` set `self.A_mat = None` in `__init__` and never reassigned it, so the `if self.A_mat is None` branch always won and the GCN was fed an identity adjacency mat. The kNN "similar-cluster" graph (which is the whole point of GRASP) was dead code.

The kNN branch also hardcoded `n_neighbors=20` while the graph has only `cluster_num` nodes (default 2), so it would have raised `n_neighbors > n_samples` had it ever run. The dead-code guard prevented the crash-prone call from ever executing.

__Fix__
Build the kNN graph unconditionally, capping neighbors to the node count with `k = min(20, cluster_num - 1)`, and keep identity only as a genuine fallback when there are too few clusters (`k < 1`). Removed the unused `self.A_mat` (the graph is rebuilt each forward from the current batch's cluster centers, so it can't be cached anyway).

__Tests__
Added a regression test that looks at the adjacency passed to the GCN. it must be a real kNN graph (off-diagonal edges, no self-loops) for `cluster_num >= 2`, and identity only for `cluster_num == 1`. Fails on the old identity-only behavior, passes on the fix.

### DIFF
--- a/pyhealth/models/grasp.py
+++ b/pyhealth/models/grasp.py
@@ -254,8 +254,6 @@ class GRASPLayer(nn.Module):
         self.GCN.initialize_parameters()
         self.GCN_2 = GraphConvolution(self.hidden_dim, self.hidden_dim, bias=True)
         self.GCN_2.initialize_parameters()
-        self.A_mat = None
-
         self.bn = nn.BatchNorm1d(self.hidden_dim)
 
     def sample_gumbel(self, shape, eps=1e-20):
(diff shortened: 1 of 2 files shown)


lehendo (approved):
Looks good to me!

DarylOkeke:
The kNN graph now runs, but the adjacency passed to the GCN is missing the self-connections from GRASP. The paper augments the graph as `A_hat = A + I` before graph convolution; this code uses `include_self=False` and passes `A` directly, and the new test requires a zero diagonal. That makes each center aggregate only neighbors (with the default two clusters, each center gets only the other center). Could we add the identity before the GCN and update the test? Paper: https://cdn.aaai.org/ojs/16152/16152-13-19646-1-2-20210518.pdf

lehendo:
@fbonc pls address Daryl's comment.

fbonc:
> The kNN graph now runs, but the adjacency passed to the GCN is missing the self-connections from GRASP. The paper augments the graph as `A_hat = A + I` before graph convolution; this code uses `include_self=False` and passes `A` directly, and the new test requires a zero diagonal. That makes each center aggregate only neighbors (with the default two clusters, each center gets only the other center). Could we add the identity before the GCN and update the test? Paper: https://cdn.aaai.org/ojs/16152/16152-13-19646-1-2-20210518.pdf

Fixed. implementation now adds the identity matrix to the kNN adjacency before it is passed to both GCN layers. also updated the regression test to assert self-connections on the diagonal while retaining the check that neighbor edges are present.

DarylOkeke:
Looks good now. The kNN graph includes the self-connections required by the paper.

fbonc:
@lehendo good to merge?

lehendo (approved):
Looks good to me!

PR #1196, opened 2026-08-20
Fix sign toggle in removal-based interpretability metrics

DarylOkeke:
`RemovalBasedMetric.compute()` aliased `y_probs` instead of copying it, so the in-place negation of NEGATIVE-class samples re-flipped the sign on every percentage in the loop — the score at 20% comes out 0.0113 with `percentages=[20]` but 1.0088 with `[10, 20]`

clone `y_probs` and `ablated_probs` before negating (the negation itself is correct, it just needed a copy), plus a test that a percentage's score doesn't depend on its position in the list. Fails at 0.997 absolute difference without the fix

only affects filters that emit `SampleClass.NEGATIVE`, so the default `threshold_sample_filter` is fine but `examples/interpretability/custom_sample_filter.py` isn't

note: could also hoist the negation out of the loop since it doesn't depend on the percentage, kept it to the two clones to stay small

### DIFF
--- a/docs/api/metrics/pyhealth.metrics.interpretability.rst
+++ b/docs/api/metrics/pyhealth.metrics.interpretability.rst
@@ -24,6 +24,11 @@ Functional API
 Removal-Based Metrics
 ---------------------
 
+For binary classifiers, a sample filter can mark class-0 predictions as
+``SampleClass.NEGATIVE``. Removal-based metrics then score probability changes
+from the class-0 perspective. Each percentage is evaluated independently, so a
+percentage's score does not depend on the other requested percentages.
+
 Base Class
 ^^^^^^^^^^
 
--- a/examples/interpretability/custom_sample_filter.py
+++ b/examples/interpretability/custom_sample_filter.py
@@ -4,7 +4,7 @@ and sufficiency metrics.
 This example demonstrates:
 1. Loading a pre-trained StageNet model with processors and MIMIC-IV dataset
 2. Computing attributions with various interpretability methods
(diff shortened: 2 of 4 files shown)


lehendo (approved):
Not approved actually. Pls look at the comments below.

lehendo (changes requested):
Actually wait, can you make an extra change one sec.

Can you just add this to your PR:

neg_mask = evaluated_drops < 0 inside the if debug and return_per_percentage: block (~line 496) reuses the neg_mask name from line 402, which is read again on the next loop iteration (lines 457/460).

since if/for don't create new scopes, debug=True with multiple percentages would silently clobber the real mask mid-loop.

lehendo:
Also unsure why your PR isn't triggering the PR checks. Make sure your branch is updated and all of that.

DarylOkeke:
> Actually wait, can you make an extra change one sec.
> 
> Can you just add this to your PR:
> 
> neg_mask = evaluated_drops < 0 inside the if debug and return_per_percentage: block (~line 496) reuses the neg_mask name from line 402, which is read again on the next loop iteration (lines 457/460).
> 
> since if/for don't create new scopes, debug=True with multiple percentages would silently clobber the real mask mid-loop.

Fixed.

fbonc:
looks good.

just make sure PR contribution rules pass.

lehendo (changes requested):
Looks good to me as well, but make sure the PR checks pass.

lehendo (approved):

PR #1199, opened 2026-08-25
Fix fairness metrics NaN error

lehendo:
0/0 produces nan silently (the intended ValueError guard checks == 0, which nan never equals), poisoning any downstream average across folds/seeds


lehendo (commented):
Looks good to me.

fbonc:
Looks good.

jhnwu3 (approved):
lgtm

PR #1200, opened 2026-08-25
Fix broken example scripts and fairness metrics import

lehendo:
- Re-enable fairness_metrics_fn import in pyhealth.metrics (was commented out); add it to __all__
- Fix SyntaxError in examples/benchmark_perf/loc/minimal_los.py from a concatenated def/class statement
- Fix readmission_mimic3_fairness.py: add missing __main__ guard (was hanging under multiprocessing), update Transformer(...) call to the current dataset-only constructor API, and fix the wrong label key
- Fix sensitive_attributes_from_patient_ids to use the current get_patient()/event-based demographics API instead of the removed dataset.patients attribute; add a verified doctest

### DIFF
--- a/docs/api/metrics/pyhealth.metrics.fairness.rst
+++ b/docs/api/metrics/pyhealth.metrics.fairness.rst
@@ -1,6 +1,10 @@
 ﻿pyhealth.metrics.fairness
 ===================================
 
+``fairness_metrics_fn`` is also importable directly from the top-level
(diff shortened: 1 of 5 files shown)


lehendo (commented):
Looks good to me.

fbonc:
Looks good

jhnwu3 (approved):
lgtm

Issue #1201, opened 2026-08-25
PyHealth 2.0: several modules still target the 1.x dataset API


AxelNoun:
## Summary

While preparing a fix for #952 I audited the modules that still reference the
PyHealth 1.x dataset API. `kg_emb` is not an isolated case. It is simply the
only one whose **import** breaks, which is why it is the only one with an issue.
The others fail later, and quietly, because the 1.x base classes now resolve to
deprecation stubs whose `__init__` only emits a warning:

```python
class BaseEHRDataset:
    """This class is deprecated and should not be used."""

    def __init__(self, *args, **kwargs):
        import warnings
        warnings.warn(
            "The BaseEHRDataset class is deprecated and will be removed in a future version.",
            DeprecationWarning,
        )
```

A subclass of such a stub constructs successfully, sets none of its attributes,
and fails somewhere downstream, often several calls later, with an error that
does not point at the cause. All of the modules below are still exported from
their packages, so users can import and instantiate them today.

This is a tracker, not a request for one large PR. I am opening it so the scope
is visible in one place; I am not proposing to fix all of it myself.

### Group A: reachable runtime failures

| Module | 1.x symbol | Breaks at | Why |
|---|---|---|---|
| `medcode/.../kg_emb` | `SampleBaseDataset` | import | `SampleKGDataset` subclassed a name removed without a stub, and its constructor passed a list into the 2.0 streaming `SampleDataset`. Addressed in #1202. |
| `datasets/mimicextract.py` | `BaseEHRDataset`, `Visit`, `parallel_apply` | runtime | The stub constructor only warns, so `parse_basic_info` fails before any of the four `parallel_apply` calls is reached. `Visit` is a stub too, and 2.0 `Patient` has no `add_visit`. |
| `datasets/shhs.py` | `BaseSignalDataset` | runtime | No local `__init__`, so the stub never sets `root`. |
| `datasets/isruc.py` | `BaseSignalDataset` | runtime | Same as SHHS. |
| `datasets/cardiology.py` | `BaseSignalDataset` | runtime (partial) | Reassigns `self.root` itself after the no-op `super().__init__`, so construction survives; the 1.x `set_task` pipeline it then relies on is gone. |
| `tasks/drug_recommendation.py` | `Visit`, `patient[i]` | runtime | `for i in range(len(patient))` followed by `visit: Visit = patient[i]`; 2.0 `Patient` defines neither `__len__` nor `__getitem__`. |
[...]

PR #1202, opened 2026-08-25
fix(kg_emb): migrate SampleKGDataset off the removed 1.x dataset API

### LINKED ISSUE #952: fix: kg_emb broken import in PyHealth 2.0

AxelNoun:
Closes #952. Supersedes #1192. Part of #1201.

## What the rename does not reach

#1192 renames `SampleBaseDataset` to `SampleDataset` across 7 files. The rename
clears the `ImportError`, but the module stays unusable: since 2.0,
`SampleDataset` is a `litdata.StreamingDataset` whose constructor expects a
directory containing `schema.pkl`, while `SampleKGDataset.__init__` still passes
a list of samples, an argument the rename does not touch.

```
>>> SampleKGDataset(samples=[{"triple": (0, 0, 1)}], entity2id={}, relation2id={})
ValueError: dir_path must be either a string, Path, or Dir, got: <class 'list'>
  litdata.streaming.resolver._resolve_dir
```

`self.samples` is then never assigned, so `__getitem__` and `stat()` raise
`AttributeError`, and `BaseKGDataset.set_task()` cannot return anything usable.
The failure moves from import time to call time.

To be fair to the author: #952 explicitly asked for the rename across eight
files, and #1192 does exactly that. The gap is in how the issue was framed.

## Root cause

`SampleKGDataset` inherited from `SampleBaseDataset` to reuse `__len__` and
`.samples`. It required none of the parent's behavioural contract and used none
of the services the hierarchy exists to provide: a knowledge graph has no
feature schema, no processors and no patient/visit index. The inheritance was
convenience reuse, not an is-a relationship.

The 2.0 migration did not introduce this. It removed the coincidence that made
it harmless, namely the overlap between the interface actually used and the one
the parent exposed. Adapting `SampleKGDataset` to the streaming contract would
mean satisfying a constraint the module has no use for, so removing the
inheritance is the minimal fix rather than the drastic one.

## What this PR does

`SampleKGDataset` becomes a standalone `torch.utils.data.Dataset`.

Models depend on a structural `typing.Protocol` (PEP 544) instead of a concrete
class. `KGEBaseModel.__init__` only ever reads `entity_num`, `relation_num` and
`task_spec_param`, so both the old and the new annotation over-specified the
contract, and both were wrong: `SampleDataset` has neither attribute.
`KGDatasetProtocol` states the capability actually required. The annotation
becomes checkable by mypy, the model layer becomes testable with lightweight
doubles, and `kg_emb` stops being coupled to changes in the main pipeline. That
last point is the durable one: a rename would have held until the next
`SampleDataset` refactor.

Also in scope:

- `split()` no longer overwrites NumPy's global random state, and validates its
  input with `ValueError` rather than `assert`, which is stripped under
  `python -O` and should not be load-bearing for user input validation. Note:
  for a given `seed=`, this returns a different partition than the previous
  implementation (different RNG algorithm). There is no prior release of this
  module, so no existing caller depends on the old ordering.
- The four models' `__main__` examples imported `SampleKGDataset` from
  `pyhealth.datasets`, where it has never existed. They now use
  `torch.utils.data.DataLoader` with `collate_fn_dict_with_padding` directly,
  since `get_dataloader` calls `set_shuffle` and is streaming-only.
- The `pandarallel` import, of a package never declared in `pyproject.toml`, is
  removed rather than added to the dependencies, since no `parallel_apply`
  exists anywhere in `kg_emb`. There is no lockfile entry to regenerate.
- `base_kg_dataset.py` and `umls.py` imported their siblings through the
  absolute package path, so the package only initialised correctly for one
  ordering of the imports in `__init__.py`, which nothing enforced. Enabling
  ruff's isort rule would have sorted `base_kg_dataset` first and broken it.

## Tests

`tests/core/test_kg_emb.py`, 22 behavioural tests: construction, indexing,
vocabularies, cardinality validation, split partitioning and reproducibility,
generic collation of variable-length ground truths, `BaseKGDataset.set_task()`
on a synthetic graph, and scoring invariants (DistMult symmetry in head and
tail, TransE margin on an exact triple).

No test asserts on a type annotation: annotations are metadata, and under
PEP 563 such an assertion compares against a string.

Every test was validated by fault injection. Reverting the fix it protects makes
it fail:

| Test | Mutation | Caught |
|---|---|:---:|
| `test_construction_and_length` | restore `super().__init__(samples, dataset_name, task_name)` | yes |
| `test_is_a_map_style_dataset` | add a fake `set_shuffle` | yes |
| `test_set_task_on_a_synthetic_graph` | drop `**kwargs` from the `SampleKGDataset(...)` call | yes |
| `test_global_numpy_state_is_untouched` | restore `np.random.seed` plus `shuffle` | yes |
| `test_variable_length_ground_truth_stays_a_python_list` | tensorise `ground_truth_head` in `__getitem__` | yes |
| ... | *(22/22, full table on request)* | yes |

These are hand-seeded mutants aimed at this PR's changes, not a systematic
mutation run, so 22/22 is not a mutation score in the usual sense. It supports a
weaker and sufficient claim: a rename-only fix cannot make this suite go green.

## Scope and limits

- This decouples `kg_emb` from the 2.0 pipeline; it does not migrate it into
  the streaming pipeline. If you would rather see the latter, the dataset layer
  here is isolated and tested and would be the starting point.
- Correctness of the plumbing, not empirical validation: no MRR or Hits@k
  compared against the original papers. Verified on synthetic graphs only, as
  UMLS requires a licence.
- `get_dataloader` remains streaming-only. `kg_emb` no longer depends on it, so
  this is not blocking here; tracked in #1201 together with the other
  modules still targeting the 1.x API.

Credit to @userjuma for the original diagnosis in #1192. The import chain
described there is accurate and was the starting point for this work.

### DIFF
--- a/docs/api/medcode.rst
+++ b/docs/api/medcode.rst
@@ -97,6 +97,50 @@ Medication codes:
     :undoc-members:
     :show-inheritance:
 
+Knowledge graph embeddings
+--------------------------
+
+``pyhealth.medcode.pretrained_embeddings.kg_emb`` trains TransE, RotatE,
+DistMult and ComplEx on an in-memory list of triples. Since PyHealth 2.0
+the sample dataset is a map-style :class:`torch.utils.data.Dataset`. Build
+the loader with :class:`torch.utils.data.DataLoader` and
+:func:`pyhealth.datasets.collate_fn_dict_with_padding` --
+:func:`pyhealth.datasets.get_dataloader` is streaming-only and calls
+``set_shuffle()``.
+
+See ``examples/kg_emb_sample_dataset.py`` for a self-contained walk-through.
+
+.. autoclass:: pyhealth.medcode.pretrained_embeddings.kg_emb.datasets.SampleKGDataset
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+.. autofunction:: pyhealth.medcode.pretrained_embeddings.kg_emb.datasets.split
+
+.. autoclass:: pyhealth.medcode.pretrained_embeddings.kg_emb.models.KGEBaseModel
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+.. autoclass:: pyhealth.medcode.pretrained_embeddings.kg_emb.models.TransE
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+.. autoclass:: pyhealth.medcode.pretrained_embeddings.kg_emb.models.RotatE
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+.. autoclass:: pyhealth.medcode.pretrained_embeddings.kg_emb.models.DistMult
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+.. autoclass:: pyhealth.medcode.pretrained_embeddings.kg_emb.models.ComplEx
+    :members:
+    :undoc-members:
+    :show-inheritance:
 
 
 
--- a/examples/kg_emb_sample_dataset.py
+++ b/examples/kg_emb_sample_dataset.py
@@ -1,0 +1,58 @@
+"""Train a TransE model on a synthetic knowledge-graph sample dataset.
+
+This example does not download UMLS. It shows the 2.0-safe path:
+
+1. Build an in-memory :class:`SampleKGDataset` (not ``SampleDataset``).
+2. Split into train/val/test folds with :func:`split`.
+3. Wrap the train fold in ``torch.utils.data.DataLoader`` using
+   :func:`collate_fn_dict_with_padding`. Do not call ``get_dataloader``:
+   that helper requires ``litdata.StreamingDataset.set_shuffle()``.
+"""
+
+import torch
+from torch.utils.data import DataLoader
+
+from pyhealth.datasets import collate_fn_dict_with_padding
+from pyhealth.medcode.pretrained_embeddings.kg_emb.datasets import (
+    SampleKGDataset,
+    split,
+)
+from pyhealth.medcode.pretrained_embeddings.kg_emb.models import TransE
+
+samples = [
+    {
+        "triple": (i % 5, i % 2, (i + 1) % 5),
+        "ground_truth_head": [i % 5, (i + 1) % 5],
+        "ground_truth_tail": [(i + 1) % 5],
+        "subsampling_weight": torch.tensor([0.25]),
+    }
+    for i in range(10)
+]
+
+dataset = SampleKGDataset(
+    samples=samples,
+    dataset_name="toy",
+    task_name="link_prediction",
+    entity2id={"a": 0, "b": 1, "c": 2, "d": 3, "e": 4},
+    relation2id={"treats": 0, "causes": 1},
+    negative_sampling=4,
+)
+print(dataset)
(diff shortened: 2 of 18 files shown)


AxelNoun:
Update: after discussing direction with @jhnwu3, the target is to keep
`SampleKGDataset` inside the `SampleDataset` hierarchy rather than decoupling from it,
with in-memory loading as an interim step for negative sampling. Moving this to draft
while I rework it, and suggesting #1192 merge first as the unblocking fix for #952.

Carrying over regardless of the base class: the `split()` fixes, the constructor
raising `AttributeError` on its own default `entity2id=None`, `stat()` returning
`None`, the phantom `SampleKGDataset` import in the models' `__main__` blocks, and the
test suite. Not carrying over: `KGDatasetProtocol` and the standalone `Dataset`.

AxelNoun:
### Architecture & Implementation Plan Update

Just dropping a quick update here to keep a clear record of the architectural decisions we aligned on via Discord, which will guide the next commits for #1202:

**1. Base Architecture (In-Memory)**
As discussed, `SampleKGDataset` will inherit directly from `InMemorySampleDataset`. This keeps us correctly under the `SampleDataset` umbrella while cleanly sidestepping the immediate complexities of handling negative sampling in a pure streaming setup. 

**2. Data Shape & Serialization (The "Tensor Trick")**
To handle the KG triples and the variable-length entity lists (`ground_truth_head` / `ground_truth_tail`), we are going to avoid slow JSON/Pickle serialization. I dug into the `tuple_time_text_processor.py` and will adapt their pattern to keep the `litdata` backend happy:
*   I will implement a custom `KGProcessor`.
*   **`fit()` phase:** Calculate the global `max_length` of the entity lists.
*   **`process()` phase:** Pad the variable-length lists and convert everything directly into pure PyTorch tensors before serialization.

**Next Steps:**
I will be working on drafting the `KGProcessor` and the base class integration tonight. I'll push the new commits once the core pipeline is running so we have concrete code to review!

AxelNoun:
###  CI/CD Green & Ready for Review

Everything is fully green! I've completed the implementation of the architectural plan we discussed:
* Reverted `SampleKGDataset` to correctly inherit from `InMemorySampleDataset`.
* Integrated the `KGProcessor` (using the "Tensor Trick" with independent max-length padding) to bypass the `litdata` serialization bottlenecks.
* Ensured negative sampling correctness in `kg_base.py` by safely reconstructing the unpadded entity lists via attention masks prior to `np.in1d` filtering.
* Passed all local repo linting rules (Ruff Python 3.13 typing updates and newly required doctests).

The PR is out of draft and ready for your final review whenever you have time!

PR #1204, opened 2026-08-25
Fix PatientLinkageMIMIC3Task crashing on invalid processor names

lehendo:
- input_schema used "integer"/"string"/"datetime" processor-type strings, none of which are registered in pyhealth.processors, so dataset.set_task() failed immediately with ValueError: Unknown processor. Fixed to use "raw" (pass-through), the correct processor for these metadata fields
- Fix the class docstring's example, which referenced a non-existent patient_linkage_mimic3 importable name instead of the class itself
- Add the missing docs page for PatientLinkageMIMIC3Task
- examples/patient_linkage_mimic3_medlink.py uses this exact class and was broken by the same bug plus two unrelated .stat()/.stats() typos; fix the typos and confirm data loading + task application now run and produce real samples

### DIFF
--- a/docs/api/tasks.rst
+++ b/docs/api/tasks.rst
@@ -219,7 +219,8 @@ Available Tasks
     MPF Clinical Prediction (FHIR) <tasks/pyhealth.tasks.mpf_clinical_prediction>
     Mortality Prediction (Next Visit) <tasks/pyhealth.tasks.mortality_prediction>
     Mortality Prediction (StageNet MIMIC-IV) <tasks/pyhealth.tasks.mortality_prediction_stagenet_mimic4>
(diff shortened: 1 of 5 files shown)


DarylOkeke:
`SampleDataset` doesn't have a `.samples` attribute, so the updated example still crashes here. `convert_to_ir_format()` only iterates its input, so passing `sample_dataset` directly should work. Could the test cover that step too?

lehendo:
@DarylOkeke fixed. Pls let me know if we are good to merge.

fbonc:
Looks good.

DarylOkeke:
> @DarylOkeke fixed. Pls let me know if we are good to merge.

Looks good!!

jhnwu3 (approved):
Thanks for the cleanup

PR #1205, opened 2026-08-25
Fix post-outcome data leakage in StageNet mortality/LOS tasks

lehendo:
- MortalityPredictionStageNetMIMIC4: exclude diagnosis/procedure codes for the admission that ends in death, and cap that admission's labs to the first 48 hours after admission instead of through discharge. Earlier, already-resolved admissions are unaffected
- LengthOfStayStageNetMIMIC4: same treatment for the target admission (the one whose LOS is the label)

### DIFF
--- a/docs/api/tasks.rst
+++ b/docs/api/tasks.rst
@@ -215,6 +215,7 @@ Available Tasks
     Drug Recommendation <tasks/pyhealth.tasks.drug_recommendation>
     EHR Generation <tasks/pyhealth.tasks.generate_ehr>
     Length of Stay Prediction <tasks/pyhealth.tasks.length_of_stay_prediction>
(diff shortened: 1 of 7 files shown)


DarylOkeke:
I think this still leaks the label through preprocessing. `is_terminal_admission` comes directly from `hospital_expire_flag`, then death cases lose final-admission ICD codes and get only 48 hours of labs, while survivors keep final-admission ICD codes and labs through discharge. So two otherwise identical patients get different feature windows based on the answer. Could the same target admission and observation window be applied to both classes? The current survivor test seems to lock in this asymmetry.

lehendo:
@DarylOkeke , fixed. Pls review.

DarylOkeke:
Looks good now. Both classes use the same target-admission window, so the label-dependent input difference is gone.

jhnwu3 (approved):
Thanks for the clean catch.

PR #1206, opened 2026-08-25
Fix broken GAMENet example; flag legacy LOC scripts as 1.1.6-only

lehendo:
- drug_recommendation_mimic4_gamenet.py called set_task() with a dead legacy function and used several other stale pre-2.0 API calls (root=/tables= kwargs, .stat()/.info(), hardcoded device="cuda"). Rewrote it to use DrugRecommendationMIMIC4, current dataset kwargs, and .stats(); dropped the CUDA hardcode so it runs on CPU too

- minimal_legacy_*.py (LOC benchmark scripts) are deliberately pre-2.0-API by design, for comparison against their 2.0-style siblings, but unlike their legacy_ver/ counterparts, they didn't disclose that. Added the same "requires pyhealth==1.1.6" note; no behavior change


lehendo (commented):
Looks good to me.

fbonc:
Looks good.

jhnwu3 (approved):
That's a good idea.

jhnwu3:
one conflict here.

PR #1208, opened 2026-08-27
Fix CNN tensor multihot input

fbonc:
**Issue**

CNN’s one-dimensional convolution path expects embedded features shaped [batch, sequence_length, embedding_dim]. Sequence processors produce that shape, but MultiHotProcessor and one-dimensional TensorProcessor inputs produce one embedding per sample, shaped [batch, embedding_dim]. Because the sequence axis is absent, CNN rejects these otherwise-supported inputs as two-dimensional tensors.
 
**Fix**

For the Conv1d path, normalize a [batch, embedding_dim] embedding to [batch, 1, embedding_dim] before validating and permuting dimensions. The existing permutation then gives Conv1d its required [batch, embedding_dim, 1] input. Inputs that already contain a sequence dimension are unchanged.

**Notes**

The regression test runs CNN with both a multi-hot feature and a one-dimensional tensor feature and verifies that forward and backward propagation complete successfully.

### DIFF
--- a/pyhealth/models/cnn.py
+++ b/pyhealth/models/cnn.py
@@ -329,6 +329,12 @@ class CNN(BaseModel):
                 x = x.to(self.device)
 
             spatial_dim = self.feature_conv_dims[feature_key]
+            # Some spatial_dim==1 features embed to [batch, embedding_dim] with no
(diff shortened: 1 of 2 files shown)


lehendo (approved):
Looks good to me!

PR #1209, opened 2026-08-27
Fix SDOH dataclass init

fbonc:
**Issue**

SdohClassifier was decorated with @dataclass while subclassing BaseModel, which is an nn.Module. The dataclass-generated \_\_init__ replaces BaseModel.\_\_init__ and never calls nn.Module.\_\_init__, so the module's internal state (_parameters, _modules, _buffers) is never created. As a result the object cannot behave as a torch module: parameters(), .to(), .eval(), and submodule assignment all raise, and @dataclass also makes the instance unhashable. The class's own docstring example (SdohClassifier()) yields an unusable object.

**Fix**

Removed the @dataclass decorator and the field(...) declarations, replacing them with an explicit \_\_init__ that calls super().\_\_init__(dataset=None) and assigns api_key, base_model_id, and adapter_model_id. Passing dataset=None is correct here because this is an LLM-based classifier that is not trained against a SampleDataset; BaseModel handles a falsy dataset by leaving feature/label keys empty. Removed the now-unused dataclass imports.

**Notes**

Added regression test test_is_initialized_nn_module in tests/core/test_sdoh.py asserting nn.Module state is initialized.

### DIFF
--- a/pyhealth/models/sdoh.py
+++ b/pyhealth/models/sdoh.py
@@ -69,7 +69,7 @@ class SdohClassifier(BaseModel):
 
     def __init__(
         self,
-        api_key: str = None,
+        api_key: str | None = None,
         base_model_id: str = 'meta-llama/Llama-3.1-8B-Instruct',
         adapter_model_id: str = 'plandes/sdoh-llama-3-1-8b',
     ):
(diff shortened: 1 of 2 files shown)


lehendo (approved):
Looks good to me!

PR #1210, opened 2026-08-27
Remove dead pre-2.0 VAE and Graph_TorchvisionModel

fbonc:
**Issue**

VAE (vae.py) and Graph_TorchvisionModel (graph_torchvision_model.py) call BaseModel.\_\_init__ with the pre-2.0 signature (dataset, feature_keys, label_key, mode), but the current BaseModel.\_\_init__ accepts only dataset. Instantiating either class therefore always raises TypeError, so both are unusable. They are pre-2.0 dead code: beyond the constructor, Graph_TorchvisionModel also calls get_label_tokenizer(), get_output_size(tokenizer), and prepare_labels(), none of which exist on the current BaseModel. Both classes are exported from pyhealth.models but referenced nowhere in the library or test suite.

**Fix**

Removed the two dead modules (pyhealth/models/vae.py, pyhealth/models/graph_torchvision_model.py) and their exports in pyhealth/models/\_\_init__.py. Also removed the now-dangling VAE autodoc page (docs/api/models/pyhealth.models.VAE.rst) and its toctree entry in docs/api/models.rst, which would otherwise break the docs build.

**Notes**

Two example files still reference the removed classes: examples/cxr/chextXray_image_generation_VAE.py and examples/graph_torchvision_model.ipynb. They were already non-functional (their models could never be constructed) and were left in place to keep this pr scoped to the dead library code. Also historical mentions in docs/log.rst (changelog) and docs/tutorials.rst (example listing) werEeleft untouched as they do not break the build.


DarylOkeke:
These removals leave two tracked examples importing names that no longer exist: `examples/cxr/chextXray_image_generation_VAE.py` imports `VAE`, and `examples/graph_torchvision_model.ipynb` imports `Graph_TorchvisionModel`. They may already have failed later during construction, but after this PR they fail immediately on import and still advertise removed APIs. Could those examples be removed or updated as part of the cleanup?

fbonc:
> These removals leave two tracked examples importing names that no longer exist: `examples/cxr/chextXray_image_generation_VAE.py` imports `VAE`, and `examples/graph_torchvision_model.ipynb` imports `Graph_TorchvisionModel`. They may already have failed later during construction, but after this PR they fail immediately on import and still advertise removed APIs. Could those examples be removed or updated as part of the cleanup?

done

lehendo (commented):
Instead of just outright deleting these models, why don't we find a way to make these models 2.0 compatible.

PR #1211, opened 2026-08-27
Fix SampleDataset subset mappings

DarylOkeke:
`SampleDataset.subset()` kept the parent dataset’s patient and record mappings, leaving stale or out-of-range indices after splitting. This rebuilds both mappings with subset-local indices for disk-backed and in-memory datasets. A regression test covers reordered and nested subsets.


fbonc:
looks good!

lehendo (approved):
Looks good to me!

PR #1213, opened 2026-08-27
Fix BIOT channel embedding ignoring emb_size

fbonc:
**Issue**

BIOTEncoder created its channel-token embedding as nn.Embedding(n_channels, 256) (biot.py:161), hardcoding dimension 256 while emb_size is user-configurable. In forward the channel-token embedding is added to the emb_size-dimensional spectral embedding, so any emb_size != 256 raises a shape-mismatch RuntimeError.

**Fix**

Changed the channel-token embedding to nn.Embedding(n_channels, emb_size) so it matches the rest of the encoder.

**Notes**

Added regression test test_model_non_default_emb_size in tests/core/test_biot.py using emb_size=128 (asserts the channel-token dim and runs a forward/backward pass). Full BIOT test suite (9 tests) passes.

### DIFF
--- a/pyhealth/models/biot.py
+++ b/pyhealth/models/biot.py
@@ -132,6 +132,7 @@ class BIOTEncoder(nn.Module):
     """Encoder for multichannel biosignals.
 
     Examples:
+        >>> _ = _get_linear_attention_transformer()
         >>> encoder = BIOTEncoder(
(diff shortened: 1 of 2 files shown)


lehendo (approved):
Looks good to me!

PR #1214, opened 2026-08-27
Fix MoleRec fallback predictor never training

fbonc:
**Issue**

On MoleRec's fallback path (no valid SMILES data), the linear fallback predictor simple_predictor was created lazily inside forward() (molerec.py:814-819). The standard PyHealth training pattern builds the optimizer from model.parameters() before the first forward pass, so the predictor's parameters were never registered with the optimizer and never trained.

**Fix**

Create simple_predictor in __init__ when has_valid_smiles is False, sized nn.Linear(hidden_dim * 2, label_size) to match the concatenated condition/procedure patient embedding. Removed the lazy creation (and its now-unnecessary .to(self.device)) from forward. Parameters are now registered before the optimizer is built.

**Notes**

Added regression test test_fallback_predictor_registered_and_trains in tests/core/test_molerec.py: it asserts simple_predictor is in named_parameters before any forward and that an optimizer step actually updates its weights.

### DIFF
--- a/pyhealth/models/molerec.py
+++ b/pyhealth/models/molerec.py
@@ -628,7 +628,14 @@ class MoleRec(BaseModel):
         
         # check if we have valid molecular data (not just dummy data)
         self.has_valid_smiles = len(self.substructure_smiles) > 1
-        
+
+        # Fallback predictor used when there is no valid molecular data. It must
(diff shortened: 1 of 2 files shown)


lehendo (approved):
Looks good to me!

PR #1215, opened 2026-08-27
Fix ECE confidence binary crash

fbonc:
**Issue**

ece_confidence_binary indexed prob[:,0] and label[:,0] (calibration.py:150), requiring 2D arrays. Its only real caller, binary_metrics_fn (binary.py:91), passes 1D positive-class probabilities and 1D 0/1 labels, so ECE and ECE_adapt always raised IndexError for binary tasks via the documented API. As a secondary defect, even for 2D input the code used column 0 (class 0) despite the docstring stating the metric is computed on class 1.

**Fix**

Compute confidence from the positive class (prob[:,1] when 2D, else the 1D array) and use the 0/1 label as the accuracy target (label[:,1] when 2D one-hot, else the 1D array). This makes the function work with the 1D shapes the caller actually uses and corrects the class-0/class-1 indexing. ece_classwise is unaffected (it calls a separate internal helper, not this function).

**Notes**

Added new test file tests/core/test_calibration_binary_ece.py with two tests. No prior tests existed for this function.

### DIFF
--- a/pyhealth/metrics/calibration.py
+++ b/pyhealth/metrics/calibration.py
@@ -149,13 +149,10 @@ def ece_confidence_binary(prob:np.ndarray, label:np.ndarray, bins=20, adaptive=F
 
     prob = np.asarray(prob)
     label = np.asarray(label)
-    # Confidence in the positive class (class 1). Accept either a 1D array of
-    # positive-class probabilities (as binary_metrics_fn passes) or a 2D
-    # (N, C) probability matrix.
+
     conf = prob[:, 1] if prob.ndim > 1 else prob
-    # Whether each sample truly belongs to the positive class. Accept either a
-    # 1D 0/1 label array or a 2D one-hot label matrix.
     acc = label[:, 1] if label.ndim > 1 else label
+    
     df = pd.DataFrame({'acc': acc, 'conf': conf})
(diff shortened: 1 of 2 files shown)


DarylOkeke:
I think this still crashes with PyHealth's normal binary model output. `BinaryLabelProcessor.process()` returns shape `(1,)`, so after batching and inference both `prob` and `label` are `(N, 1)`. Since `ndim > 1`, this code tries `[:, 1]`, but column 1 does not exist. Could the shape handling distinguish flat `(N,)`, single-column `(N, 1)` (use column 0), and two-column `(N, 2)` (use positive-class column 1), with a regression test for `(N, 1)`?

fbonc:
> I think this still crashes with PyHealth's normal binary model output. `BinaryLabelProcessor.process()` returns shape `(1,)`, so after batching and inference both `prob` and `label` are `(N, 1)`. Since `ndim > 1`, this code tries `[:, 1]`, but column 1 does not exist. Could the shape handling distinguish flat `(N,)`, single-column `(N, 1)` (use column 0), and two-column `(N, 2)` (use positive-class column 1), with a regression test for `(N, 1)`?

fixed!

lehendo (approved):
Looks good to me!

PR #1216, opened 2026-08-27
Handle two-dimensional time embeddings

fbonc:
**Issue**

SinusoidalTimeEmbedding divided its frequency indices by half - 1. For dim=2, half is 1, so the only frequency was computed as 0/0 and every embedding contained NaN values.

**Fix**

Clamp to at least 1. This gives the single-frequency case a finite frequency of 1 while leaving the frequency schedule unchanged for all dimensions greater than 2.

**Notes**

Added regression tests tests/core/test_unified_embedding.py with finite-output coverage for dim=2 and a known-value check for dim=6.


lehendo (approved):
Looks good to me!

PR #1217, opened 2026-08-29
Fix calibration self-inclusion leak in NeighborhoodLabel (NCP)

lehendo:
calibrate() queried sklearn's NearestNeighbors.kneighbors() with an explicit X argument equal to the fitted calibration set, which (unlike the implicit no-argument form) returns each point as its own nearest neighbor at distance 0


DarylOkeke:
This leave-one-out change breaks the alpha search -- without each calibration point in its own neighborhood, some points can still fail at alpha_tilde=0, so the method may never reach the target coverage. The commit’s 23/30 case shows this. The authors’ implementation avoids self-inclusion with a separate tuning set. Use separate tuning and calibration sets instead of training the test model until coverage becomes achievable.

PR #1221, opened 2026-08-30
ClusterLabel audit + tests

lehendo:
Title is self explanatory.


fbonc:
The documentation promises coverage for each cluster separately, but the clusters are built using the same calibration patients that are later used to set each cluster's threshold. That means the calibration data helps define the groups and choose their thresholds, so the usual finite-sample guarantee does not apply.

Fit K-means on training embeddings only, then assign calibration patients with `predict()`. The test should also measure coverage inside each cluster, rather than only across the whole test set.

lehendo:
@fbonc , fixed. pls review.

DarylOkeke:
The new coverage test still averages all test patients together, so one well-covered cluster can hide another that under-covers. Please calculate and check coverage separately for each cluster.

PR #1222, opened 2026-08-30
Fix medlink collate_fn: robust to heterogeneous per-sample keys

lehendo:
collate_fn built its output keys from only the first sample in a batch. If a batch mixed samples that had a mined hard negative (s_n) with samples that didn't, this either crashed with KeyError('s_n') or silently produced a shorter, misaligned list for that field.

Fix: union keys across all samples, and drop s_n for the whole batch if any sample in it is missing one.


fbonc:
looks good.

jhnwu3 (approved):
lgtm

PR #1223, opened 2026-08-31
Fix under-coverage in CovariateLabel's finite-sample correction

lehendo:
test_forward_with_embeddings_uses_per_point_weight failed on CI (0.3241... == 0.3241...). Root cause: a test point's KDE-derived weight is always a small fraction of total calibration weight, so whether it shifts the selected quantile index is essentially a coin flip depending on random model init

### DIFF
--- a/docs/api/calib.rst
+++ b/docs/api/calib.rst
@@ -42,6 +42,25 @@ New to calibration and uncertainty quantification? Check out this complete examp
 
 This example shows the complete pipeline from model training to uncertainty-aware predictions with formal coverage guarantees.
 
+.. note::
+
+   ``CovariateLabel``'s finite-sample correction implements Corollary 1 of
+   Tibshirani, Barber, Candes, and Ramdas, "Conformal Prediction Under
+   Covariate Shift" (NeurIPS 2019, https://arxiv.org/abs/1904.06019): the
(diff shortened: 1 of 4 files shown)


fbonc:
The threshold should be calculated separately for each test patient, using that patient's embedding. However, `Trainer.inference()` only passes the data batch to the model, not `test_embeddings`. Since the example uses `Trainer.inference()`, it always falls back to one shared, approximate threshold for every patient.

I think put the embeddings in each batch or compute them inside `CovariateLabel.forward()`. Also raise an error when the number of embeddings does not match the batch size instead of ignoring extras.

DarylOkeke:
Another boundary issue -- Corollary 1 includes scores equal to the threshold. Since larger scores should be included, this should use >= instead of >.. Tied scores at the threshold are excluded and can under-cover. Switch it to `>=` and add a tie test

lehendo:
@fbonc @DarylOkeke , fixed. pls review again.

DarylOkeke:
The standalone example uses the new per-patient threshold path, but the main class example and in-file demo still use the old inference path. That gives everyone one approximate threshold, so those examples should use the per-patient path too.

fbonc:
@lehendo what he said

PR #1224, opened 2026-09-01
Fix: Use configured TransformerDeID field names

fbonc:
**Issue**
TransformerDeID.deidentify() always called the model with text= and labels=, even when the dataset configured different feature and label field names. Custom schemas therefore failed with a missing-key error

**Fix**
Build the de-identification call from self.feature_key and self.label_key so inference follows the field names established from the dataset schema

**Notes**
Regression test added


lehendo (approved):
Looks good to me!

PR #1227, opened 2026-09-02
Remove module-level RNG reset from gnn.py

lehendo:
pyhealth/models/gnn.py called torch.manual_seed(3) and np.random.seed(1) at module level. Since pyhealth/models/init.py imports from this file, simply running import pyhealth.models silently overwrote any seed a caller had already set.

The fix was to remove both module-level seed calls, which has been done.


fbonc:
looks good.

PR #1228, opened 2026-09-05
Fix SampleDataset cache cleanup

DarylOkeke:
Closing a SampleDataset could delete reusable task cache files. This tracks whether the dataset owns its directory, so close() only removes temporary data and keeps persistent or shared caches intact.

### DIFF
--- a/docs/api/datasets.rst
+++ b/docs/api/datasets.rst
@@ -82,6 +82,9 @@ future runs without re-fitting.
 Calling ``samples.subset(...)`` rebuilds both lookups with indices local to the
 new dataset, so they remain valid after repeated splitting.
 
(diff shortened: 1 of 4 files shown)


fbonc:
in `tests/core/test_caching.py:172`, the test suite fails cause this test still expects the processed sample cache from `BaseDataset.set_task()` to be deleted when its context exits.

this PR intentionally preserves that reusable cache, so `index.json` and `schema.pkl` still exist. fix is to update these assertions to expect both files to remain.

fbonc:
otherwise looks good.

lehendo:
@DarylOkeke , can you address the failing CI.

PR #1230, opened 2026-09-05
Fix T1D DKA timing

DarylOkeke:
DKA events before a patient's Type 1 diabetes diagnosis were being treated as future outcomes because the time difference used an absolute value. This keeps the date order, excludes earlier DKA events, and adds boundary tests for the 90-day window.


fbonc:
looks good.

lehendo (approved):
Looks good to me!

lehendo (approved):
Looks good to me!

PR #1233, opened 2026-09-09
Add NestedMultiHotProcessor and vectorise HALO's visit encoding

janezdu:
EHR generation feeds HALO a nested list of per-visit codes. The existing NestedSequenceProcessor emits code indices *padded to the longest visit seen* during `it, so one outlier sets the width for the whole dataset: on eICU a single visit holds 3,951 code entries (the same diagnosis re-charted through a stay) while a typical visit holds about five. HALO then unpacked that back into multi-hot vectors with a triple-nested Python loop.

Add **NestedMultiHotProcessor** (registered as `nested_multihot`), which emits one multi-hot row per visit, sized by the vocabulary rather than by the worst-case visit -- 8.6x smaller per patient on a 921-code eICU vocabulary (4.5 KB vs 38.7 KB). Repeats within a visit collapse to a single 1, which is what set-membership models already did with the index form.

Switch EHRGeneration to it and rewrite `HALO._encode_visits` as a vectorised placement into the context window. The old loop cost an `.item()` per patient -- a CUDA sync each -- and a single-element kernel launch per code, which measured at 84.5% of a training step at batch 128 on an A100 (13.887 s encode vs 2.553 s forward+backward), and ~99.8% in steady state once CUDA warmup is excluded.

decode_dataset is updated to invert the new encoding: reading multi-hot rows as indices would see only 0s and 1s and decode every visit as empty.

<pad> (0) and <unk> (1) keep their indices, so the vocabulary is interchangeable between the two processors.

Tests: test_nested_multihot_processor.py covers the processor directly, and test_halo_encode_equivalence.py asserts HALO sees identical tensors from either processor, so the switch changes no results.

---

**tl;dr** NestedMultiHotProcessor + vectorised HALO._encode_visits. 8.6× less memory per patient, and encoding drops from 84.5% of a training step to negligible. Equivalence test proves results don't change. 10 files, +472/−33.


tests/core/test_halo_encode_equivalence.py
```diff
@@ -0,0 +1,112 @@
...
+import unittest
+
+import torch
+
+from pyhealth.datasets import create_sample_dataset, get_dataloader
+from pyhealth.models import HALO
+from pyhealth.processors import NestedSequenceProcessor
+
+
+def legacy_encode(cfg, visits, device):
```
jhnwu3: Do we need to have a legacy check here? I think it's ok to just use our patched/fixed generative modeling setup here.

jhnwu3 (changes requested):
Tldr; We may not need test_halo_encode_equivalence.py here. It seems redundant here.

janezdu:
Noticed the GPT2 which used to use **NestedSequenceProcessor** is now taking in the NestedMultiHotProcessor and converting it back into the sequence. Changed it so it just takes NestedSequenceProcessor as before.

Also bundled in new tasks for MedGAN/CorGAN which are bag-of-codes, since they don't have a task yet. Result:

| Model | Task | Processor |
|---|---|---|
| **HALO** | `VisitMultiHotGeneration` | `NestedMultiHotProcessor` |
| **GPT2** | `VisitSequenceGeneration` | `NestedSequenceProcessor` |
| **PromptEHR** | `VisitSequenceGeneration` | `NestedSequenceProcessor` |
| **MedGAN** | `PatientCodeSetGeneration` | `MultiHotProcessor` |
| **CorGAN** | `PatientCodeSetGeneration` | `MultiHotProcessor` |

note now the task `EHRGeneration` is an abstract superclass so it errors out; anyone calling it directly needs to pick one of the three subclasses. EHRGenerationMIMIC3/4 are unchanged.

pyhealth/tasks/generate_ehr.py
```diff
@@ -63,51 +77,94 @@
...
+        if event_type is not None:
+            self.event_type = event_type
+        if code_attr is not None:
+            self.code_attr = code_attr
+        if min_visits is not None:
+            self.min_visits = min_visits
+
+    def _visits(self, patient: Patient) -> list[list[str]]:
+        """Ordered per-admission code lists, empty admissions dropped."""
+        visits: list[list[str]] = []
```
jhnwu3: My only concern here is that this type of parent class only works on MIMIC and not other datasets like eICU, so should anyone try to extend this parent class. It would fail. 

I think it would be good to simplify and simply just have a flat list of different task classes for each model/dataset. We could maybe rename all of this to the MIMIC-series of generative tasks here and get rid of the inheritance as it's unnecessary.

examples/halo_mimic3.py
```diff
@@ -2,19 +2,23 @@
...
 This example demonstrates:
 1. Loading MIMIC-III data
-2. Applying the EHRGenerationMIMIC3 task (per-visit ICD-9 code sequences)
-3. Creating a SampleDataset with a NestedSequenceProcessor
+2. Applying the EHRGenerationMIMIC3 task (per-visit ICD-9 code sets,
+   multi-hot -- the encoding HALO consumes; GPT2/PromptEHR instead take
+   VisitSequenceGeneration, and MedGAN/CorGAN PatientCodeSetGeneration)
+3. Creating a SampleDataset with a NestedMultiHotProcessor
 4. Training the HALO generator with its custom training loop
 5. Generating synthetic patients
```
jhnwu3, line 10: If you could vibe generate more examples using the revamped tasks, that would be awesome!

jhnwu3 (changes requested):

janezdu:
Flattened, got rid of EHRGeneration. now everything is EHR{generation,codeset}{mimic3,mimic4} and I'll add eICU later.

jhnwu3 (approved):
lgtm

PR #1234, opened 2026-09-09
Fix MICRON handling of unequal code lists

fbonc:
**Issue**
MICRON treats individual codes as visits. Unequal diagnosis and procedure counts can crash concatenation, while padding alone can cause predictions to ignore the shorter list.

**Fix**
Pool flat code lists into one visit, preserve nested visit histories, and return zero reconstruction loss for single-visit inputs.

**Notes**
Regression tests verify that both features receive prediction gradients and nested inputs retain their visits.


lehendo:
@fbonc , can you take care of the checks.

MateehUllah (approved):
Reviewed the current two-file patch. Pooling flat SequenceProcessor inputs into a single visit fixes the diagnosis/procedure length mismatch without changing nested-sequence visit structure, and collapsing the corresponding mask to one visit keeps it aligned with the pooled embedding. Returning a scalar zero reconstruction loss for a one-visit tensor is also appropriate because there is no inter-visit transition to reconstruct; the BCE term still supplies the training signal. The added tests cover gradient flow through both unequal flat features and preservation of nested visits. I did not find a blocking correctness issue.

PR #1237, opened 2026-09-11
Fix bare except in ranking metrics

francesca-crespi:
Replaced the bare except: clause with except ImportError: in metrics/ranking.py so that only missing pytrec_eval imports are caught


jhnwu3:
CI might flag this incorrectly (we'll probably need a smarter interface later for catching missing documentation), but this is a nice change. Will check back in when test cases finish.

